PR #32, opened 2018-06-01
Explaining the return in case of exception.

FernandoBittencourt:
Explaining the return in case of an exception.
To get to the return at the end of the method is only for this case. 
Thus avoiding an ignored exception.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptor.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptor.java
@@ -79,9 +79,8 @@ public class FeignContentGzipEncodingInterceptor extends BaseRequestInterceptor 
 			final long length = Long.parseLong(strLen);
 			return length > getProperties().getMinRequestSize();
 		} catch (NumberFormatException ex) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@FernandoBittencourt Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-openfeign&pullRequestId=32)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-openfeign&pullRequestId=32) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

ryanjbaxter:
Not sure how this really makes a difference in the functionality, can you elaborate?

FernandoBittencourt:
The reason for my suggestion to change the code is improove to read the code because it explicitly states that 'return false' is belong to the exception and thus eliminates the need for a comment explaining that this exception intentionally not contain any handling.
This contributes to the evolution of the code, leaving explicitly that there are no other flows that go through that return.

pivotal-issuemaster:
@FernandoBittencourt Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-openfeign&pullRequestId=32)!

PR #40, opened 2018-06-27
expose configuration metadata for property 'feign.hystrix.enabled'

mbenson:


ryanjbaxter:
I dont recall us adding this file anywhere else, and we have it documented.  I can see that this would add additional information in the IDE.  @spencergibb thoughts?

PR #50, opened 2018-07-28
Support to set loadBalancerKey

ileler:
Support to override CachingSpringLoadBalancerFactory to customize the FeignLoadBalancer instance.
use example:
```
@Configuration
public class FeignClientConfiguration {

    @Bean
    @Primary
    public Client feignClient(SpringClientFactory clientFactory, @Autowired(required = false) LoadBalancedRetryFactory retryFactory) {
        return new LoadBalancerFeignClient(new Client.Default(null, null),
                new CustomCachingSpringLoadBalancerFactory(clientFactory, retryFactory), clientFactory);
    }

    class CustomCachingSpringLoadBalancerFactory extends CachingSpringLoadBalancerFactory {

        private volatile Map<String, FeignLoadBalancer> cache = new ConcurrentReferenceHashMap<>();

        public CustomCachingSpringLoadBalancerFactory(SpringClientFactory factory) {
            super(factory);
        }

        public CustomCachingSpringLoadBalancerFactory(SpringClientFactory factory, LoadBalancedRetryFactory loadBalancedRetryPolicyFactory) {
            super(factory, loadBalancedRetryPolicyFactory);
        }

        @Override
        public FeignLoadBalancer create(String clientName) {
            FeignLoadBalancer client = this.cache.get(clientName);
            if(client != null) {
                return client;
            }
            IClientConfig config = this.factory.getClientConfig(clientName);
            ILoadBalancer lb = this.factory.getLoadBalancer(clientName);
            ServerIntrospector serverIntrospector = this.factory.getInstance(clientName, ServerIntrospector.class);
            client = loadBalancedRetryFactory != null ? new RetryableFeignLoadBalancer(lb, config, serverIntrospector,
                    loadBalancedRetryFactory) {
                protected void customizeLoadBalancerCommandBuilder(final FeignLoadBalancer.RibbonRequest request, final IClientConfig config,
                                                                   final LoadBalancerCommand.Builder<FeignLoadBalancer.RibbonResponse> builder) {
                    builder.withServerLocator(request.getRequest().headers().get(Constants.HEADER_CLIENT_IP));
                }
            } : new FeignLoadBalancer(lb, config, serverIntrospector) {
                protected void customizeLoadBalancerCommandBuilder(final FeignLoadBalancer.RibbonRequest request, final IClientConfig config,
                                                                   final LoadBalancerCommand.Builder<FeignLoadBalancer.RibbonResponse> builder) {
                    builder.withServerLocator(request.getRequest().headers().get(Constants.HEADER_CLIENT_IP));
                }
            };
            this.cache.put(clientName, client);
            return client;
        }

    }

}
```

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/CachingSpringLoadBalancerFactory.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/CachingSpringLoadBalancerFactory.java
@@ -36,8 +36,8 @@ import com.netflix.loadbalancer.ILoadBalancer;
  */
 public class CachingSpringLoadBalancerFactory {
 
-	private final SpringClientFactory factory;
-	private LoadBalancedRetryFactory loadBalancedRetryFactory = null;
+	protected final SpringClientFactory factory;
+	protected LoadBalancedRetryFactory loadBalancedRetryFactory = null;
 
 	private volatile Map<String, FeignLoadBalancer> cache = new ConcurrentReferenceHashMap<>();
 
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/FeignLoadBalancer.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/FeignLoadBalancer.java
@@ -118,7 +118,7 @@ public class FeignLoadBalancer extends
 		private final Request request;
 		private final Client client;
 
-		RibbonRequest(Client client, Request request, URI uri) {
+		protected RibbonRequest(Client client, Request request, URI uri) {
 			this.client = client;
 			setUri(uri);
 			this.request = toRequest(request);
@@ -170,6 +170,13 @@ public class FeignLoadBalancer extends
 			};
 		}
 
+		public Request getRequest() {
+			return request;
+		}
+
+		public Client getClient() {
+			return client;
+		}
 
 		@Override
 		public Object clone() {
@@ -182,7 +189,7 @@ public class FeignLoadBalancer extends
 		private final URI uri;
 		private final Response response;
 
-		RibbonResponse(URI uri, Response response) {
+		protected RibbonResponse(URI uri, Response response) {
 			this.uri = uri;
 			this.response = response;
 		}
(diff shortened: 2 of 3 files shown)


marcingrzejszczak:
Why do you want to do that? What problem are you trying to solve? You have written no tests.

marcingrzejszczak:
What are those specific requirements?

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/CachingSpringLoadBalancerFactory.java
```diff
@@ -36,8 +36,8 @@
  */
 public class CachingSpringLoadBalancerFactory {
 
-	private final SpringClientFactory factory;
-	private LoadBalancedRetryFactory loadBalancedRetryFactory = null;
+	protected final SpringClientFactory factory;
+	protected LoadBalancedRetryFactory loadBalancedRetryFactory = null;
```
adriancole: this probably should have been final, looking at the code

adriancole: strictly speaking you don't need to expose these. your subtype can override and shadow the fields, or make similarly named ones.

adriancole:
please change your example to actually use the DebugRule you mentioned. Right now, the two examples are not connected.

Also, please first try a composition approach. See if you can use the existing code to form what you need vs inheritance. For example, see if you can forward calls to existing types somehow, without subtyping.

Anyway, if you can't please copy/paste the least code possible so it is easy to see the entire integration, including DebugRule. We might be able to identify another way to achieve the same goal, which doesn't require updating two projects, and exposing their constructors, to achieve.

ileler:
@adriancole Hello, I submitted a test case. Can you see my needs from here? If this PR does not pass, my needs seem to have no other way to achieve it.My requirement is actually to set the `loadBalancerKey` by overriding `FeignLoadBalancer`. Then filter the server by `loadBalancerKey` in the `IRule` instance of the `ribbon`.

ileler:
Is there anyone tracking this pr?

ryanjbaxter:
@adriancole want to take another pass at this?

ileler:
@ryanjbaxter Hey, adriancole seems to be very busy and has no time to look at this PR. Can you find other members to evaluate whether to pass this PR? We have to improve efficiency, because it is not good to have other users waiting for this small feature to be officially released!

ileler:
Well, thank you for your reply. Sorry to bother you, but this PR is still pending, and the reviewer needs to reply to confirm the status. Thank you for your comments, then please ask other members to consider whether to merge.

ryanjbaxter:
Not sure what you mean, if you have another PR please submit it and we will take a look.

Issue #62, opened 2018-08-30
Add support for MultipartFile


Hronom:
I have tested `spring-cloud-openfeign` of version `2.0.1` and it not automatically process `MultipartFile`.

This means that constructions like this:
```
@FeignClient(value = "file-service")
public interface FileServiceClient {
    @PostMapping(value = "/upload-file", consumes = MediaType.MULTIPART_FORM_DATA_VALUE, produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
    ResponseEntity<String> processFileTest(
        @RequestPart(value = "file", required = true) MultipartFile file
    );
}
```
**not works.**

To make it work - you need.
**1.**
Include in your `pom.xml` dependency:
```
        <dependency>
            <groupId>io.github.openfeign.form</groupId>
            <artifactId>feign-form-spring</artifactId>
            <version>3.3.0</version>
        </dependency>
```

**2.**
Configure you feign client with next configuration:
```
@Configuration
public class ServiceClientConfig {
    @Autowired
    private ObjectFactory<HttpMessageConverters> messageConverters;

    @Bean
    public Encoder feignFormEncoder() {
        return new SpringFormEncoder(new SpringEncoder(messageConverters));
    }
}
```

Only after that, you will be able to use `@RequestPart(value = "file", required = true) MultipartFile file,`.
This should be supported in Spring Boot and Cloud by default. 

As far as I understand this change should be added only here org.springframework.cloud.openfeign.FeignClientsConfiguration or somewhere else?

Related discussion was started here https://github.com/spring-cloud/spring-cloud-netflix/issues/867#issuecomment-416911321

antemooo:
**TL;DR** <br />**Transform your MultiPartFile to MultiValueMap. [See example that i provide on StackOverFlow ](https://stackoverflow.com/a/58641199/9159029)**

----------
The answer mentioned by [@Hronom](https://github.com/Hronom) is correct answer to have your FeignClient method signature the same as the actual endpoint signature that you are trying to call. However, there is a way around that does not require you to define a FormEncoder configuration and add any extra dependency because in some applications you are not allowed to do that (enterprise shit); all what you need is to transform your MultipartFile to a MultiValueMap and it will perfectly work as the standard encoder will be able to serialize it.

PR #66, opened 2018-09-05
Added support for MultipartFile

Hronom:
Proposed solution for https://github.com/spring-cloud/spring-cloud-openfeign/issues/62

### Notes
The solution is based on
```
        <dependency>
            <groupId>io.github.openfeign.form</groupId>
            <artifactId>feign-form-spring</artifactId>
            <version>3.3.0</version>
        </dependency>
```

Class `feign.form.spring.SpringFormEncoder` was added inside `org.springframework.cloud.openfeign.support.SpringEncoder` because it's doesn't has some checks:
1. Check of  `bodyType` on null.
`org.springframework.cloud.openfeign.support.SpringEncoder#encode` don't handle situation when `bodyType` is null.

2. No check for content type `multipart/form-data` in header. 
`feign.form.FormEncoder` requires `multipart/form-data` in header for proper processing.

All this checks added and tests are passed.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
@@ -24,6 +24,7 @@ import java.lang.reflect.Type;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Collection;
+import java.util.Objects;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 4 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -51,6 +54,8 @@
 
 	private static final Log log = LogFactory.getLog(SpringEncoder.class);
 
+	private final SpringFormEncoder springFormEncoder = new SpringFormEncoder();
```
spencergibb: `feign-form-spring` is marked as optional (though included in the starter). Either it should function as an option (which would fail here), or the optional should be removed and then can be removed from the starter.

spring-cloud-starter-openfeign/pom.xml
```diff
@@ -39,6 +39,10 @@
 			<groupId>io.github.openfeign</groupId>
 			<artifactId>feign-core</artifactId>
 		</dependency>
+		<dependency>
```
spencergibb: This is no longer needed now that it is not optional.

Hronom: Done

PR #79, opened 2018-11-08
Add support for feign's QueryMap annotation for Object mapping

momothereal:
The `@QueryMap` annotation in OpenFeign allows support for Map-to-query and Object-to-query encoding. In the current state, there's an inconsistency between OpenFeign and spring-cloud-openfeign:

* In OpenFeign, you can mark an Object parameter with `@QueryMap` to use it as query parameters instead of a body.
* In Spring, having an Object parameter in a GET mapping will map it to query automatically. Therefore, there is no equivalent annotation for `@QueryMap`.

Since `SpringMvcContract` overrides the Feign BaseContract, which handles the QueryMap annotation, having an Object parameter would pass it as a body instead of query parameters.

**Another solution** to this inconsistency would be to modify the `RequestParamParameterProcessor` to allow for Object parameters for QueryMap *if* the method  in question is a GET. Let me know if you'd rather have such a solution instead.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/annotation/QueryMapParameterProcessor.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/annotation/QueryMapParameterProcessor.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2013-2018 the original author or authors.
+ *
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
How about a test?

ryanjbaxter:
Can you add some documentation as well?

Issue #67, opened 2018-09-06
Support different custom configurations for multiple feign clients with the same name


menelaosbgr:
I have an issue that is related with: https://github.com/spring-cloud/spring-cloud-netflix/issues/1211

We have a service accepts multi-part upload requests for one method.
We had another method that was accepting a JSON object ( request body).

```
@FeignClient(name = "ab-document-store", fallback = DocumentStoreFallback.class, configuration = MultiPartSupportConfiguration.class)
public interface DocumentStoreClient {

	@Headers("Content-Type:multipart/form-data")
	@RequestMapping(path = "/document-store/v1/upload", method = RequestMethod.POST) // , consumes = MediaType.MULTIPART_FORM_DATA_VALUE, produces = MediaType.MULTIPART_FORM_DATA_VALUE)
	DocumentMetaData uploadDocument(@RequestPart("files") @Param("files") MultipartFile files);

//other methods here
		
	@RequestMapping(path = "/document-store/v1/save/as/zip")
	DocumentMetaData saveDocumentsZip(@RequestBody ZipFileModel zipFileModel);
```

For multi-part to work, we were using a configuration class and an interceptor:

```
public class MultiPartSupportConfiguration {

	@Autowired
	private ObjectFactory<HttpMessageConverters> messageConverters;

	@Bean
	public Encoder feignFormEncoder() {
		return new MultipartFormEncoder(new SpringEncoder(messageConverters));
	}

	@Bean
	public Logger.Level feignLoggerLevel() {
		return Logger.Level.HEADERS;
	}
}

public class MultipartFormEncoder extends SpringFormEncoder {

	public MultipartFormEncoder(Encoder delegate) {
		super(delegate);
	}

	@Override
	public void encode(Object object, Type bodyType, RequestTemplate template) throws EncodeException {
		if(bodyType.equals(MultipartFile.class)) {
			template.header("Content-Type", "multipart/form-data");
			super.encode(object, bodyType, template);
			return;
		}
		super.encode(object, bodyType, template);
		
	}
}
```

But, we we are seeing a problem. Spring ( or feign) is forcing requests to go through  "ReflectiveFeign$BuildFormEncodedTemplateFromArgs". 
 see: https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/ReflectiveFeign.java#L352 ). 

This is removing all variables except request params ( formVariables).

[...]

PR #90, opened 2018-12-04
Support Multiple Clients Using The Same Service

Ph0ndragX:
Fixes gh-67

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
@@ -212,6 +212,27 @@ hystrix:
           strategy: SEMAPHORE
 ----
 
+If we want to create multiple feign clients with the same name or url
+so that they would point to the same server but each with a different custom configuration then
+we have to use `contextId` attribute of the `@FeignClient` in order to avoid name
+collision of these configuration beans.
+
+[source,java,indent=0]
+----
+@FeignClient(contextId = "fooClient", name = "stores", configuration = FooConfiguration.class)
+public interface FooClient {
+    //..
+}
+----
+
+[source,java,indent=0]
+----
+@FeignClient(contextId = "barClient", name = "stores", configuration = BarConfiguration.class)
+public interface BarClient {
+    //..
+}
+----
+
 === Creating Feign Clients Manually
 
 In some cases it might be necessary to customize your Feign Clients in a way that is not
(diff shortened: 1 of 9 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
```diff
@@ -16,9 +16,11 @@
 
 package org.springframework.cloud.openfeign;
 
-import java.util.Map;
-import java.util.Objects;
-
+import feign.*;
```
ryanjbaxter: We dont use * imports.  Also the import order is wrong

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/invalid/FeignClientValidationTests.java
```diff
@@ -64,24 +68,56 @@ public void testNameAndValue() {
...
+		context.close();
+	}
+
+	@Configuration
+	@Import({FeignAutoConfiguration.class, HttpClientConfiguration.class})
+	@EnableFeignClients(clients = {DuplicatedFeignClientNamesConfiguration.FooClient.class,
+			DuplicatedFeignClientNamesConfiguration.BarClient.class})
+	protected static class DuplicatedFeignClientNamesConfiguration {
+
+		@FeignClient(name = "bar", serviceId = "bar")
```
ryanjbaxter: I have a lot of questions, but this example configuration doesnt make any sense to me, how are these different?

ryanjbaxter:
I dont think we can change the use of the `serviceId` parameter now as it was deprecated, this could break people.

@spencergibb thoughts?

docs/src/main/asciidoc/spring-cloud-openfeign.adoc
```diff
@@ -73,6 +73,8 @@ in your external configuration (see
 
 A central concept in Spring Cloud's Feign support is that of the named client. Each feign client is part of an ensemble of components that work together to contact a remote server on demand, and the ensemble has a name that you give it as an application developer using the `@FeignClient` annotation. Spring Cloud creates a new ensemble as an
 `ApplicationContext` on demand for each named client using `FeignClientsConfiguration`. This contains (amongst other things) an `feign.Decoder`, a `feign.Encoder`, and a `feign.Contract`.
+It is possible to override the name of that ensemble by using the `contextId`
```
ryanjbaxter: I think an example of what this enables would be useful

PR #93, opened 2018-12-07
 Add feign-mock dependency

jawobar:
Please consider adding feign-mock to dependencies, so it can be picked up when using BOM. I am using it in my tests and need to specify version manually.

https://github.com/OpenFeign/feign/tree/master/mock

### DIFF
--- a/spring-cloud-openfeign-dependencies/pom.xml
+++ b/spring-cloud-openfeign-dependencies/pom.xml
@@ -95,6 +95,11 @@
 			</dependency>
 			<dependency>
 				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-mock</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.github.openfeign</groupId>
 				<artifactId>feign-ribbon</artifactId>
 				<version>${feign.version}</version>
 			</dependency>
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
We are not going to add dependencies we do not use.

BernhardStreitTooltime:
Is there a way to add them to the dependency management, so that I can import the dependency in my pom (and have the same version as the other feign libs), @ryanjbaxter ?

ryanjbaxter:
No we are not going to add dependencies we do not use to our dependency management

BernhardStreitTooltime:
Hm, is there then a suggested way how to solve that issue?

Now I have feign in some version x getting added via spring boot starter poms, and I manually have to add a dependency to feign-mock in my own pom that needs to be kept in sync with the feign version that  comes with the starter poms.

Remember, it's not a 3rd-party thing, it's a module of the official feign project: https://github.com/OpenFeign/feign/tree/master/mock

And it's version is kept in sync with feign itself:

```xml
  <parent>
    <artifactId>parent</artifactId>
    <version>10.3.0-SNAPSHOT</version>
  </parent>

  <artifactId>feign-mock</artifactId>
  <name>Feign Mock</name>
  <description>Feign Mock</description>
```

ryanjbaxter:
My colleague has convinced me otherwise, I added feign-mock to our dependency management section in the 2.1.x and master branches

BernhardStreitTooltime:
Thank you very much :)

Issue #85, opened 2018-11-22
Upgrade to Feign 10.1.0


OlgaMaciaszek:
There's Feign 9.7.0 in [dependencies](https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-dependencies/pom.xml#L10)

Should  be upgraded to 10.1.0.

OlgaMaciaszek:
There are issues that would make it not backward compatible, so we would have to wait with that for a major release. Below is a summary:

It looks like Feign has switched to only handling text bodies - UTF-8 charset is set by default, the `body()` method does not change the request charset anymore and there is no accessible setter to do it manually and we have methods in `SpringEncoder` that also handled binary and protobuf bodies; I'm still planning to discuss it with the OpenFeign devs, but most probably we would have to remove support for those files as well.

The functionality related to operating on uris has been changed. The method `feign.RequestTemplate#insert` that we use in `org.springframework.cloud.openfeign.support.SpringMvcContract#processAnnotationOnClass` is now deprecated and it also does a different thing that what it did before - it now sets absolute host name and not the remaining path segments -as previously-; I guess we could be able to executed the same operations on the uris using the new `uriTemplate` field, however it has private read access as of now - it has a public write access and append possibility, but since we used the previous `insert()` implementation to prepend the uris with segments from `@RequestMapping` annotations, it would not be sufficient for us - I guess public read access would have to be added for `uriTemplate` to maintain same functionality on our side. Alternatively, we could try to use `path()` to extract the segments present in `uriTemplate()` and then prepend them with segments from the annotations and set in `requestTemplate` using `uri()` method, but that would be hacky and I don't really like the idea.

The `feign.Contract.BaseContract#parseAndValidateMetadata` method has started adding query parameters when we have a `@RequestParam` `List` method parameters - we could fix it, though, by not calling the original method in our implementation and, instead, implementing that bit ourselves.

robachmann:
Thank so you much for your preliminary work, @OlgaMaciaszek. I really hope you manage to upgrade the dependency to a newer version in the near future. Maybe there will be a Feign version `10.2.0` or `11.0.0` where Spring and Feign meet in the middle with some shared concepts.
As a developer I hope being able to access `FeignException`'s content in a future version. 
Keep it up and thank you again

kdavisk6:
@OlgaMaciaszek 

[...]

PR #95, opened 2018-12-13
Upgrade openfeign to 10.1.0

OlgaMaciaszek:
Fixes gh-85


spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/SpringMvcContractTests.java
```diff
@@ -51,13 +53,12 @@
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect;
-
+import static com.google.common.base.Charsets.UTF_8;
```
spencergibb: no `com.google`.

OlgaMaciaszek: sure, wrong import. Thanks :)

PR #106, opened 2019-01-14
Upgrade feign-form to 3.5.0

### LINKED ISSUE #105: Incompatible versions between Feign and Feign-Form
The changes between Feign 10.0.1 and 10.1.0 make 10.1.0 incompatible with Feign Form 3.3.0 (or even 3.4.1) when using `FormEncoder`, resulting in multiple values for `Content-Type` in the request: see https://github.com/OpenFeign/feign-form/issues/56 and https://github.com/OpenFeign/feign/issues/866.

The dependency on `feign-form` should be bumped to 3.5.0 (which I've tested successfully with the `spring-cloud-dependencies` Greenwich.RC2 BOM by setting the `feign-form.version` property).

fpavageau:
The earlier versions are not compatible with Feign 10.1.0 when using the
form encoder.

Fixes #105

### DIFF
--- a/spring-cloud-openfeign-dependencies/pom.xml
+++ b/spring-cloud-openfeign-dependencies/pom.xml
@@ -15,7 +15,7 @@
 	<description>Spring Cloud OpenFeign Dependencies</description>
 	<properties>
 		<feign.version>10.1.0</feign.version>
-		<feign-form.version>3.3.0</feign-form.version>
+		<feign-form.version>3.5.0</feign-form.version>
 	</properties>
 	<dependencyManagement>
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
Could we perhaps add a test that would catch this in the future?

fpavageau:
@ryanjbaxter I was afraid someone would ask that. 😅 

I guess a test with a client using the `FormEncoder` from Feign Form and verifying that everything is correct in the request could be written.

ryanjbaxter:
It would be helpful if you could add a test

Issue #171, opened 2019-04-01
Creating clients with @RequestHeader in v2.1.1.RELEASE throws UnsupportedOperationException


rejierd:
When upgrading spring-cloud-openfeign from v2.1.0.M2 to v2.1.1.RELEASE in our application, we're unable to create Feign clients with the @RequestHeader.

The issue is that [this commit](https://github.com/spring-cloud/spring-cloud-openfeign/commit/a1a45125ea616663c2ec984f069fe6cc745e0892#diff-14659cdcdcb3748b828088b0a3fbae47R53) contains code that tries to add elements to an immutable collection, which then throws an uncaught UnsupportedOperationException.

I've created an MVCE [here](https://github.com/rejierd/spring-cloud-issue).

There are two workarounds: Either remove @RequestHeader or the produces argument from @RequestMapping.

OlgaMaciaszek:
Hello @rejierd . Thanks for reporting this. Will provide a fix.

PR #172, opened 2019-04-02
Fix handling adding multiple parametrs for key.

OlgaMaciaszek:
Fixes gh-171.


OlgaMaciaszek:
> Assuming the test failed before the fix, LGTM.
@spencergibb yes, when upgrading to newer version of OpenFeign,  I haven't realised that this could now return unmodifiable list.

PR #181, opened 2019-05-06
Improve pageable parameter names.

zbin:
Synchronize with the `Pageable` parameter name in `SpringDataWebProperties`

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
@@ -36,6 +36,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.data.web.SpringDataWebProperties;
 import org.springframework.boot.autoconfigure.http.HttpMessageConverters;
 import org.springframework.cloud.openfeign.support.PageJacksonModule;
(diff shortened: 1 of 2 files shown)


zbin:
This change is used to synchronize the parameter name of the `Pageable` with the settings in `SpringDataWebProperties`.

The processing logic of the current Pageable object has a fixed parameter name. Cannot be applied to some scenes that rename `pageSize` or `pageNumber`.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/PageableSpringEncoder.java
```diff
@@ -37,6 +38,21 @@
 
 	private final Encoder delegate;
 
+	/**
+	 * Page index parameter name.
+	 */
+	private String pageParameter = "page";
```
spencergibb: Add setters for parameter names.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -87,7 +91,7 @@ public Encoder feignEncoder() {
 	@ConditionalOnClass(name = "org.springframework.data.domain.Pageable")
 	@ConditionalOnMissingBean
 	public Encoder feignEncoderPageable() {
-		return new PageableSpringEncoder(new SpringEncoder(this.messageConverters));
+		return new PageableSpringEncoder(dataWebProperties, new SpringEncoder(this.messageConverters));
```
spencergibb: Don't pass as a constructor parameter, but if not null here, pass names to setters on `PageableSpringEncoder`.

zbin:
This test failed seems not my reason, I am not sure.

spencergibb:
Can you rebase?

zbin:
> Can you rebase?

@spencergibb 
Sorry, I'm not sure if this is the case.

There was a wrong  'rebase' operation before, and I performed a' push --force' to override it.
Now I try' Rebase' and keep in line with the base branch.

However, there was still a unit test error. But running it on my local host does not have this error.

Please help me.

PR #197, opened 2019-07-18
Fix FeignClientBuilder fallback

moziyan:

Issue #184, opened 2019-05-17
Customized fallback or fallbackFactory should be allowed to replace the default


lw5946:
I don't want to define a fallBack for every FeignClient.

In FeignClientFactoryBean.java, the default fallback is void.class, the default fallbackFactory is void.class.

I need a way to replace the default with a customized fallback.

luryson:
I also need to create some feign clients manually. But `FeignClientBuilder.Builder`'s methods are all not generic. Additionally, `fallback()` and `fallbackFactory()` should pass `Class<?>` maybe.

PR #215, opened 2019-09-02
FeignClientBuilder.Builder should always return the generic builder

edenhaus:
Hi,
in the current version the FeignClientBuilder.Builder is return always it's raw type and not the generic one. This leads to many unckeck cast warnings or type mismatch errors.

With the current version the following code, will produce an type mismatch error
```
@Bean
    public DummyClass dummyClass() {
        return feignClientBuilder.forType(DummyClass.class, "dummyClass")
            .url("URL").build();
    }
//Error:(25, 57) java: incompatible types: java.lang.Object cannot be converted to DummyClass
```
The problem is, that the method url is returning **Builder** instead of **Builder&lt;T&gt;**. 

I also changed the parameter for the two methods fallback & fallbackFactory to **Class&lt;?&gt;** (as mention in #184), so the code is backward-compatible. Also the BuilderTest will set the fallback to Object.
There is already a discussion about the two fallback methods (#197 and #184) and we should discuss about these methods there.
The intent of this pullrequest is only to make the builder return is generic implementation, so we can avoid casting it to the correct class.

Feel free to ask or change something in the code.
Cheers,
Robert

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientBuilder.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientBuilder.java
@@ -61,43 +61,42 @@ public class FeignClientBuilder {
 					.fallbackFactory(void.class);
 		}
 
-		public Builder url(final String url) {
+		public Builder<T> url(final String url) {
 			this.feignClientFactoryBean.setUrl(FeignClientsRegistrar.getUrl(url));
 			return this;
 		}
 
-		public Builder contextId(final String contextId) {
+		public Builder<T> contextId(final String contextId) {
 			this.feignClientFactoryBean.setContextId(contextId);
 			return this;
 		}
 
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@edenhaus Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-openfeign&pullRequestId=215)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-openfeign&pullRequestId=215) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@edenhaus Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-openfeign&pullRequestId=215)!

edenhaus:
Merged the master into this PR

PR #238, opened 2019-11-02
SpringDecoder should respect 'decode404' settings

michaldo:
'decode404' is not clearly defined, but I think that for response 404, Feign client should return `Optional.empty()` or `null`.

`Optional` case is implemented correctly, but second case was wrong - `SpringDecoder` tried construct object basing on response body instead just return `null`.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringDecoder.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringDecoder.java
@@ -24,6 +24,7 @@ import java.lang.reflect.WildcardType;
 
 import feign.FeignException;
 import feign.Response;
+import feign.Util;
 import feign.codec.DecodeException;
 import feign.codec.Decoder;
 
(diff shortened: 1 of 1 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringDecoder.java
```diff
@@ -50,6 +50,9 @@ public SpringDecoder(ObjectFactory<HttpMessageConverters> messageConverters) {
 	@Override
 	public Object decode(final Response response, Type type)
 			throws IOException, FeignException {
+		if (response.status() == 404) {
```
spencergibb: Doesn't seem right to me

michaldo: Let me explain.
Assume there is a class `Movie`
```
public class Movie {
    private String title;
    private String description;
    private String message;
}
```
and endpoint `/favouriteMovie` returns
```
{
  "title":"Bodyguard",
  "description":"Singer star hires security officer",
  "message":"Movie actually rented"
}
```
Feign client returns `Movie` with
> m.title: Bodyguard
m.description: Singer star hires security officer
m.message: Movie actually rented

But when endpoint `/favouriteMovie` returns
```
{
  "timestamp":"2019-11-02T07:14:16.894+0000",
  "status":404,
  "error":"Not Found",
  "message":"No message available",
  "path":"/"
}
```
actually Feign client returns `Movie` with
>m.title=null
m.description=null
m.message=No message available

From client perspective, such handling `decode404=true` is useless - it is hard to distinguish  correct and not-found answer.

When I was writing this response, I found a bit better implementation:
`return null` -> `return Util.emptyValueOf(type)` 
(PR is modified)

OlgaMaciaszek:
After a discussion, the team has decided not to introduce this change.

Issue #225, opened 2019-09-23
Body type is ignored in SpringEncoder


ahmadmo:
In the [`SpringEncoder`](https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java), the parameter `bodyType` is ignored and not passed to the `canWrite` and `write` methods of the [`GenericHttpMessageConverter`](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/converter/GenericHttpMessageConverter.java) ([here](https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java#L94) and [here](https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java#L112)) which is important for them to determine whether or not they support and can write the `requestBody`.
This could be done similar to [this](https://github.com/spring-projects/spring-framework/blob/master/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/AbstractMessageConverterMethodProcessor.java#L284).

BTW, as long as this problem exists, what is the alternative solution for  `GenericHttpMessageConverter`s to determine the actual `Type` of the `requestBody` for example when it is a `ParameterizedType` which cannot be determined from `requestBody.getClass()`?

ryanjbaxter:
Would you like to submit a PR to fix the problem?

ahmadmo:
@ryanjbaxter sure, in a few days.

PR #242, opened 2019-11-07
Fix bug in SpringEncoder (#225)

ahmadmo:
This commit fixes a bug in the `SpringEncoder` that causes the `bodyType` to
be ignored and not passed to generic message converters.

Fixes gh-225


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -91,9 +92,15 @@ public void encode(Object requestBody, Type bodyType, RequestTemplate request)
...
 			}
 
-			for (HttpMessageConverter<?> messageConverter : this.messageConverters
+			for (HttpMessageConverter messageConverter : this.messageConverters
 					.getObject().getConverters()) {
-				if (messageConverter.canWrite(requestType, requestContentType)) {
+				GenericHttpMessageConverter genericConverter = messageConverter instanceof GenericHttpMessageConverter
+						? (GenericHttpMessageConverter) messageConverter : null;
+				// noinspection unchecked
+				if (genericConverter != null
```
OlgaMaciaszek: Please extract these condition checks into a separate method.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -109,9 +116,16 @@ public void encode(Object requestBody, Type bodyType, RequestTemplate request)
 
 					FeignOutputMessage outputMessage = new FeignOutputMessage(request);
 					try {
-						@SuppressWarnings("unchecked")
-						HttpMessageConverter<Object> copy = (HttpMessageConverter<Object>) messageConverter;
-						copy.write(requestBody, requestContentType, outputMessage);
+						if (genericConverter != null) {
```
OlgaMaciaszek: we are checking this twice in this method. I would recommend switching it up, so we have first block, where:
1) check `genericConverter != null`
2)  `genericConverter.canWrite(bodyType, requestType,
								requestContentType)`
3) execute `genericConverter.write(requestBody, bodyType,
									requestContentType, outputMessage);`
or if it does not enter this block, then an analogical block for a `messageConverter` and then you can even extract the parts when you run `.canWrite(...)` for each of the converter into a separate method, and thus simplify the code.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -91,9 +92,15 @@ public void encode(Object requestBody, Type bodyType, RequestTemplate request)
 				}
 			}
 
-			for (HttpMessageConverter<?> messageConverter : this.messageConverters
```
OlgaMaciaszek: why are remove the generic wildcard?

ahmadmo: I thought it was unnecessary to have the generic wildcard here.
Also, having generic wildcard prevents us from calling `messageConverter.write` because of the generic parameter `T t`. hence, the only option would be casting the `messageConverter` to `HttpMessageConverter` or `HttpMessageConverter<Object>`.
So, I decided to omit the generic wildcard and use `// noinspection unchecked` instead to suppress warnings.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
ahmadmo: I agree with you. I was trying to keep my changes small and use the same style of code used [here](https://github.com/spring-projects/spring-framework/blob/master/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/AbstractMessageConverterMethodProcessor.java#L283).
I will do some refactoring based on your comment and commit my changes soon.

ahmadmo:
> The functionality looks good, but have added some comments to clean up the code a bit. Also, after you add changes, could you change the target branch from `master` to `2.1.0.x`? Like this, we will get it working also for `Greenwich.SR4`.

Should I open a new pull request or just change the base branch from `master` to `2.1.x`?

ryanjbaxter:
@ahmadmo just change the base

OlgaMaciaszek:
Changing the base back to `master`, as it would require additional work to readjust to Greenwich.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
OlgaMaciaszek: Ok.

OlgaMaciaszek:
Hi @ahmadmo due to some changes done in the release process, this PR has gone messed up (version compatibility). Will clean it up and merge.

OlgaMaciaszek:
Commits merged in https://github.com/spring-cloud/spring-cloud-openfeign/pull/272 .

ahmadmo:
> Hi @ahmadmo due to some changes done in the release process, this PR has gone messed up (version compatibility). Will clean it up and merge.

Thanks.

PR #247, opened 2019-11-13
Use the context id if present in the HystrixTargeter

### LINKED ISSUE #173: hystrix: can not get the right fallback instance
```version: 2.1.0```
```feign.hystrix.enabled: true```

```
@FeignClient(contextId = "foo-FooClient", name = "foo",
    configuration = FooClientConfiguration.class, fallback = FooClientFallback.class)
public interface FooClient {

  @GetMapping("foo")
  String foo();

}


@Configuration  // here configuration only be loaded into ApplicationContext corresponding to the contextId
public class FooClientConfiguration {

  @Bean
  public FooClientFallback fooClientFallback(){
    return new FooClientFallback();
  }

}


public class FooClientFallback implements FooClient {

  @Override
  public String foo() {
    return "foo:fallback";
  }
}
```

Caused by: java.lang.IllegalStateException: No fallback instance of type class outside.client.FooClientFallback found for feign client foo

I am trying to diagnose this problem. I find contextId attribute of the @FeignClient annotationis has changed the name of the ApplicationContext ensemble after upgrading to 2.1.0,
but the corresponding target function in  HystrixTargeter.class still use the ApplicationContext expressed by the name attribute of the @FeignClient to get fallback instance.

source code：https://github.com/spring-cloud/spring-cloud-openfeign/blob/aba2abba8a0c651e3c539123e43429bf2083bb50/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/HystrixTargeter.java#L46

 ```

@Override
	public <T> T target(FeignClientFactoryBean factory, Feign.Builder feign, FeignContext context,
						Target.HardCodedTarget<T> target) {
		if (!(feign instanceof feign.hystrix.HystrixFeign.Builder)) {
			return feign.target(target);
		}
		feign.hystrix.HystrixFeign.Builder builder = (feign.hystrix.HystrixFeign.Builder) feign;
		SetterFactory setterFactory = getOptional(factory.getName(), context,
			SetterFactory.class);
		if (setterFactory != null) {
			builder.setterFactory(setterFactory);
		}
		Class<?> fallback = factory.getFallback();
		if (fallback != void.class) { // I think 'factory.getName()' should be replaced with 'factory.getContextId' here
			return targetWithFallback(factory.getName(), context, target, builder, fallback);
		}
		Class<?> fallbackFactory = factory.getFallbackFactory();
		if (fallbackFactory != void.class) { // the same above
			return targetWithFallbackFactory(factory.getName(), context, target, builder, fallbackFactory);
		}

		return feign.target(target);
	}
```

as a contrast：

https://github.com/spring-cloud/spring-cloud-openfeign/blob/75faca6c17181e27692205c34928298de73591ea/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java#L228-L230

### here is a complete [example](https://github.com/jojo921/openfeign-issues-173) which can reproduce this problem

ryanjbaxter:
Fixes #173

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,6 @@ jobs:
     branches:
       ignore:
         - gh-pages # list of branches to ignore
-    resource_class: large
     steps:
       - checkout
       - restore_cache:
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/HystrixTargeter.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/HystrixTargeter.java
@@ -22,6 +22,8 @@ import feign.hystrix.FallbackFactory;
 import feign.hystrix.HystrixFeign;
 import feign.hystrix.SetterFactory;
 
+import org.springframework.util.StringUtils;
+
 /**
  * @author Spencer Gibb
  * @author Erik Kringen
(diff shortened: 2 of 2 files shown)


ryanjbaxter:
@jskim1991 @gdong42 @Imccccc @jojo921 could one of you verify this fix?

jskim1991:
@ryanjbaxter last night I was doing something almost exactly the same as what you have done locally. For me, it worked.

PR #250, opened 2019-11-16
Changing DefaultGzipDecoder's charset to UTF-8

jskim1991:
Problem Statement:
 - When tested with a UTF-8, DefaultGzipDecoder returned with garbage data because it is using latin-1 charsets.
 - The added test case in the PR proves such case.

Suggestion:
 - [SpringEncoder.java](https://github.com/spring-cloud/spring-cloud-openfeign/blob/1ce5ca1ba83a96036efa105d1fedfb1a72f0303f/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java#L136) uses UTF-8 by default  So it makes sense to use UTF-8 when decoding.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/DefaultGzipDecoder.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/DefaultGzipDecoder.java
@@ -20,12 +20,12 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.zip.GZIPInputStream;
 
 import feign.FeignException;
 import feign.Response;
-import feign.Util;
 import feign.codec.Decoder;
 
 import org.springframework.cloud.openfeign.encoding.HttpEncoding;
@@ -72,7 +72,7 @@ public class DefaultGzipDecoder implements Decoder {
 		try (GZIPInputStream gzipInputStream = new GZIPInputStream(
 				response.body().asInputStream());
 				BufferedReader reader = new BufferedReader(
-						new InputStreamReader(gzipInputStream, Util.ISO_8859_1))) {
+						new InputStreamReader(gzipInputStream, StandardCharsets.UTF_8))) {
 			String outputString = "";
 			String line;
 			while ((line = reader.readLine()) != null) {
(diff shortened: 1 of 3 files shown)


jskim1991:
@ryanjbaxter any feedback? If you want, I can make it into another option and users can decide which charset they want to use

ryanjbaxter:
Thanks!  I dont see why we weren't using `UTF-8` before.

@spencergibb @OlgaMaciaszek and ideas?

@jskim1991 I would submit this PR against the 2.1.x branch and we can merge it forward.

spencergibb:
I don't know, but it was intentional. Maybe have it configurable if user wants previous default?

jskim1991:
@ryanjbaxter This code was introduced in 2.2.x. So it is not possible to submit this against 2.1.x. I didn't consider much about charset when I first wrote this. However, now I see that default value should be matching with what you guys think default should be. My guess is it is UTF-8 based on code elsewhere such as SpringEncoder.

jskim1991:
@ryanjbaxter @spencergibb In terms of charset, it would depend on the language and country. As you see in the added test case in this PR, languages using UTF8 will fail. So if we want to support globally, having an option provides more variety. In terms of how to write the test case, I couldn't think of any better case that uses a different language. If you don't like that or have any suggestions, please let me know.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/DefaultGzipDecoder.java
```diff
@@ -72,7 +72,7 @@ private String decompress(Response response) throws IOException {
 		try (GZIPInputStream gzipInputStream = new GZIPInputStream(
 				response.body().asInputStream());
 				BufferedReader reader = new BufferedReader(
-						new InputStreamReader(gzipInputStream, Util.ISO_8859_1))) {
+						new InputStreamReader(gzipInputStream, StandardCharsets.UTF_8))) {
```
ryanjbaxter: Why did you use ISO_8859_1 originally?

spencergibb: :man_shrugging:

jskim1991: @ryanjbaxter @spencergibb  I originally thought latin-1 is used the most but I guess I was wrong about that.

Issue #126, opened 2019-01-31
Optional request body should be supported by feign


dimw:
In Spring MVC it is possible to [annotate body parameter as optional](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestBody.html) using `@RequestBody(required = false)` parameter annotation, e.g.:

``` java
@FeignClient("foo-service")
public interface FooService {
    @RequestMapping(value = "/foo", method = RequestMethod.POST)
    void doFoo(@RequestBody(required = false) FilterDto filter);
}
```

However, Spring-Feign contract does not consider the optionality flag which leads to an _IllegalStateException_ on proxy bootstrapping in consumer when calling something like:

``` java
fooService.doFoo(null);
```

```
java.lang.IllegalArgumentException: Body parameter 0 was null
    at feign.Util.checkArgument(Util.java:102)
    at feign.ReflectiveFeign$BuildEncodedTemplateFromArgs.resolve(ReflectiveFeign.java:323)
    at feign.ReflectiveFeign$BuildTemplateByResolvingArgs.create(ReflectiveFeign.java:213)
    at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:72)
    at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:103)
```

It is expected that `null` can be passed as parameter since it is optional.

Using:
- spring-cloud-starter-feign:1.1.0.RELEASE
- feign-core:8.16.2

PR #263, opened 2019-12-12
Adds support for optional request body Fixes gh-126

matt62king:
Adds support for ```@RequestBody(required = false)```

When request body is optional and null, encoder will add an empty body by default.

Fixes gh-126


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -147,6 +150,11 @@ else if (messageConverter instanceof ProtobufHttpMessageConverter
 			}
 			throw new EncodeException(message);
 		}
+		else {
+			if (OPTIONAL_REQUEST_BODY.equals(request.bodyTemplate())) {
```
spencergibb: Is there any way to do this without a fake template body?

matt62king: Not one that I can see at the request level.

I think it could be done ate the Client level, not sure if we would want to go that broad for this.

OlgaMaciaszek: The empty template is sort of a workaround, indeed. If that was the only option, I would probably still like to go with this, however, this PR was closed for lack of tests and feedback and not because the functionality proposal was rejected. If that gets merged, we could then reuse it on Contract level, without the workaround. @matt62king let me know, if you would like to try ammending that OpenFeign PR and getting it merged? If not, I can work on it with the OpenFeign team.

matt62king: @OlgaMaciaszek I don't think I will be working the OpenFeign PR. I have not worked with that project before.

OlgaMaciaszek: ok @matt62king . Will try to submit sth like that to them.

OlgaMaciaszek:
Have created another issue in OpenFeign: https://github.com/OpenFeign/feign/issues/1383.

Issue #257, opened 2019-12-05
Suggest open the FeignBlockingLoadBalancerClient


echooymxq:
In some case, we always want to extends the loadBalancerClient to do some magic.
But the `FeignBlockingLoadBalancerClient` is not public.

fanticat:
YES，we have the same demands，if we can config the SSlContext to disable hostname verification, some times the SSL hosts‘s certificate is self signed

OlgaMaciaszek:
I'm not sure we want to make it public. Providing an interface and a non- public implementation allows the users to plug in custom implementations while at the same time allows us to change the class if necessary. We can still consider it, we would have to have more people requesting that however.

In the meantime, I'm thinking we could allow some additional config options. Could you link to code snippets with your custom implementations and I will try to make it work for you via config options?

echooymxq:
@OlgaMaciaszek Thanks for your reply, eg: In our project, in this part([SeataFeignObjectWrapper](https://github.com/alibaba/spring-cloud-alibaba/blob/1d99e3597789b3e7aea9881e1faabead8aab2157/spring-cloud-alibaba-seata/src/main/java/com/alibaba/cloud/seata/feign/SeataFeignObjectWrapper.java#L52-L62)), we rewritten the `FeignBlockingLoadBalancerClient` ([SeataFeignBlockingLoadBalancerClient](https://github.com/alibaba/spring-cloud-alibaba/blob/1d99e3597789b3e7aea9881e1faabead8aab2157/spring-cloud-alibaba-seata/src/main/java/com/alibaba/cloud/seata/feign/SeataFeignBlockingLoadBalancerClient.java#L78)) to adapt the new loadblancer feign client.

OlgaMaciaszek:
Hi all, thanks for the feedback. After all, I will be making it public as a part of the fix of: https://github.com/spring-cloud/spring-cloud-openfeign/issues/259 .

PR #264, opened 2019-12-16
Use delegate if absolute url provided.

OlgaMaciaszek:
Fixes gh-259
Fixes gh-257


Been24:
@OlgaMaciaszek 
as the title ,Use delegate if absolute url provided.
but if delegate Is still  a LoadBalancer ,should we unwrap until not? 
Like this:
image
wish your help

Issue #124, opened 2017-04-05
Passing decode404 configuration globally using Configuration


tanmaybinaykiya:
Is there a way to pass the decode404 property to all FeignClients by instantiating a bean through a configuration class, like it is possible for ErrorDecoder or Contract etc?

OlgaMaciaszek:
@tanmaybinaykiya @matt62king as discussed in https://github.com/spring-cloud/spring-cloud-openfeign/pull/267 what is the use case where that makes more sense as opposed to using the existing `decode404` config property?

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

tanmaybinaykiya:
It's been 3 years since I created the issue. I can't remember what the use case was.

PR #267, opened 2019-12-21
Docode404 bean configuration

matt62king:
Adds the ability to set decode 404 globally by providing a bean in configuration. 

Defaults to false

Fixes gh-124


matt62king:
I think there is already a configuration property, and was asking to set via a bean in a configuration class. 

I could go back to the issue and see if we can determine a use case. I really can't think of one unless you wanted to set the decode404 conditionally for some reason.

matt62king:
Closing this because the reporter can not remember the use case.

PR #270, opened 2020-01-01
Upgrade to 10.7.0.

OlgaMaciaszek:


OlgaMaciaszek:
Note: this PR goes back to using deprecated `Request.create(...)` and `request.charset()` methods, because the non-deprecated methods now require passing the `requestTemplate` and [the only method that allows retrieving it from currently existing request](https://github.com/OpenFeign/feign/blob/9584dfafc647159f5503e895603d48c5ba126e82/core/src/main/java/feign/Request.java#L349-L351) is annotated as `@Experimental`.

OlgaMaciaszek:
Fixes gh-269

Issue #240, opened 2019-11-07
Make interface Targeter public


jacek-kaczmarek:
Problem description:
The type `org.springframework.cloud.openfeign.Targeter` is package-private. However:
* it's defined as @ConditionalOnMissingBean (suggesting it could be replaced by spring-cloud users),
* its initial commit message for type `org.springframework.cloud.openfeign.Targeter` reveals the intention to allow users to implement Feign clients without using Ribbon.

Request:
Could this interface be made public?
Now that Ribbon is deprecated, there might be more interest in this type, e.g. for lightweight service location.

Workaround:
defining `Targeter` implementations and `@Configuration` classes in the interface package works but is not elegant.

echooymxq:
The request is reasonable.

spencergibb:
PRs welcome

PR #281, opened 2020-01-19
changed Targeter to public fixes gh-240

matt62king:
Quick one here. Seems like it has been popping up a lot. 

fixes gh-240


matt62king:
Closing the PR because it duplicated here https://github.com/spring-cloud/spring-cloud-openfeign/pull/282

Issue #244, opened 2019-11-08
Support customizing Feign.Builder via a FeignBuilderCustomizer


aaronjwhiteside:
Enhancement

There are some aspects of the `Feign.Builder` which are currently not customizable out of the box, such as https://github.com/spring-cloud/spring-cloud-openfeign/issues/243. 

I want to voice a suggestion that a new interface `FeignBuilderCustomizer` be added, in the same vein as many Spring Boot `*Customizers`, that would allow people to arbitary make changes to the `Builder` before it is used. If any new features are added to OpenFeign but are not yet supported by Spring Cloud OpenFeign this would allow people to workaround those limitations, at least until whatever is lacking is supported in a more direct manor. Or perhaps not for whatever the reason may be, but it would still allows people to implement their own workarounds more easily in that case.

PR #289, opened 2020-01-30
add customizer for feign builder

matt62king:
Adds a ```Customizer``` for ```Feign.Builder``` to allow customization. 

When used ```FeignBuilderCustomizer``` will override other configurations.

Effects ```@FeignClient``` and ```FeignClientBuilder``` building.

fixes gh-244

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignBuilderCustomizer.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignBuilderCustomizer.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.openfeign;
+
+import feign.Feign;
+
+import org.springframework.security.config.Customizer;
+
+/**
+ * Allows application to customize the Feign builder.
+ *
+ * @author Matt King
+ */
+public interface FeignBuilderCustomizer extends Customizer<Feign.Builder> {
+
+}
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignBuilderCustomizer.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignBuilderCustomizer.java
@@ -18,13 +18,14 @@ package org.springframework.cloud.openfeign;
 
 import feign.Feign;
 
-import org.springframework.security.config.Customizer;
-
 /**
  * Allows application to customize the Feign builder.
  *
  * @author Matt King
  */
-public interface FeignBuilderCustomizer extends Customizer<Feign.Builder> {
+@FunctionalInterface
+public interface FeignBuilderCustomizer {
+
+	void customize(Feign.Builder builder);
 
 }
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
@@ -99,12 +99,21 @@ class FeignClientFactoryBean
 		// @formatter:on
 
 		configureFeign(context, builder);
-		context.getInstances(contextId, FeignBuilderCustomizer.class).values().stream()
-				.sorted(AnnotationAwareOrderComparator.INSTANCE)
-				.forEach(feignBuilderCustomizer -> feignBuilderCustomizer
-						.customize(builder));
+		applyBuildCustomizers(context, builder);
 
 		return builder;
+	}
+
+	private void applyBuildCustomizers(FeignContext context, Feign.Builder builder) {
+		Map<String, FeignBuilderCustomizer> customizerMap = context
+				.getInstances(contextId, FeignBuilderCustomizer.class);
+
+		if (customizerMap != null) {
+			customizerMap.values().stream()
+					.sorted(AnnotationAwareOrderComparator.INSTANCE)
+					.forEach(feignBuilderCustomizer -> feignBuilderCustomizer
+							.customize(builder));
+		}
 	}
 
 	protected void configureFeign(FeignContext context, Feign.Builder builder) {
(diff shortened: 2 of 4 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignBuilderCustomizer.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.openfeign;
+
+import feign.Feign;
+
+import org.springframework.security.config.Customizer;
```
aaronjwhiteside: Pretty sure we don't want a dependency on Spring Security to be able to customise Feign.Builder?

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
```diff
@@ -84,6 +84,7 @@ public void afterPropertiesSet() throws Exception {
 	}
 
 	protected Feign.Builder feign(FeignContext context) {
+		FeignBuilderCustomizer customizer = get(context, FeignBuilderCustomizer.class);
```
aaronjwhiteside: It would be good if we also supported more than a single Customizer a List<> that honours @Order etc..

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignBuilderCustomizer.java
matt62king: Removed

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
matt62king: Removed

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
```diff
@@ -84,6 +95,8 @@ public void afterPropertiesSet() throws Exception {
 	}
 
 	protected Feign.Builder feign(FeignContext context) {
+		Map<String, FeignBuilderCustomizer> customizerMap = context
```
aaronjwhiteside: I think this is redundant? Since you get the map of customizers in the `getOrderedCustomizers()` method again..

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
```diff
@@ -97,10 +110,43 @@ public void afterPropertiesSet() throws Exception {
...
 	}
 
+	private List<FeignBuilderCustomizer> getOrderedCustomizers(FeignContext context) {
+		Map<String, FeignBuilderCustomizer> customizerMap = context
+				.getInstances(contextId, FeignBuilderCustomizer.class);
+		BeanDefinitionRegistry registry = (BeanDefinitionRegistry) applicationContext
+				.getAutowireCapableBeanFactory();
+
+		List<FeignBuilderCustomizer> customizers = customizerMap.keySet().stream()
+				.sorted(Comparator.comparing(o -> getOrderedValue(o, registry)))
```
aaronjwhiteside: implementing a manual `getOrderedValue()` isn't necessary, have a look at https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/core/annotation/AnnotationAwareOrderComparator.html

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
matt62king: Left this one and removed the whole ```getOrderedCustomizers()``` method

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
matt62king: That is much cleaner, Thank you

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
aaronjwhiteside: now that you got rid of the method, you can also get rid of the intermediate variable..

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
```diff
@@ -97,6 +101,9 @@ public void afterPropertiesSet() throws Exception {
 		// @formatter:on
 
 		configureFeign(context, builder);
+		customizerMap.values().stream().sorted(AnnotationAwareOrderComparator.INSTANCE)
+				.forEach(feignBuilderCustomizer -> feignBuilderCustomizer
```
aaronjwhiteside: 👍

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
matt62king: removed

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -138,6 +138,13 @@ public FeignLoggerFactory feignLoggerFactory() {
 		return new DefaultFeignLoggerFactory(this.logger);
 	}
 
+	@Bean
+	@ConditionalOnMissingBean(FeignBuilderCustomizer.class)
+	public FeignBuilderCustomizer feignBuilderCustomizer() {
```
IlyasYOY: Hello! Nice work done here!

I'm curious why you need this bean declaration here if it's just an empty lambda?

matt62king: To ensure at least one instance of ```FeignBuilderCustomizer``` so there is no worry of null pointer errors when using ```FeignContext.getInstances()```

aaronjwhiteside: I didn't notice this before, I assumed this was part of the test code, but now that I do it seems very suspicious to me.

I personally would prefer the code in `FeignClientFactoryBean` be robust in the face of nulls.

matt62king: Removed empty implementation and added null checks

OlgaMaciaszek:
Fixes gh-244.

PR #296, opened 2020-02-10
added support to disable feign parent configurations, fixes gh-132

matt62king:
Adds support to disable feign parent configurations.

Can be controlled with the ```inheritParentContext``` parameter on ```@FeignClient```.

fixes gh-132


spencergibb:
I wonder if there is a way to do this without adding a new attribute on `@FeignClient`. Maybe an optional FeignClientConfigurer bean with the new boolean as well as the `primary` boolean from the annotation. That one could be deprecated and removed in 3.0.0. This could go in the 2.2.x branch as well.

OlgaMaciaszek:
@matt62king thanks for the PR. After addressing @spencergibb 's comment above, could you submit it against the `2.2.x` branch so that it gets into the `Hoxton` release train?

matt62king:
@OlgaMaciaszek made changes and switched to 2.2.x. Lots of conflict. Would I pulling in things that don't belong?

Would it be better off to rebranch and cherry pick?

matt62king:
@OlgaMaciaszek Made new pr pointing to 2.2.x

https://github.com/spring-cloud/spring-cloud-openfeign/pull/326

can fix the conflicts here if one point to master is needed

spencergibb:
We merge forward

matt62king:
made new pr to 2.2.x https://github.com/spring-cloud/spring-cloud-openfeign/pull/326

PR #309, opened 2020-03-20
add support for an FeignErrorDecoderFactory

### LINKED ISSUE #308: introduce a ErrorDecoderFactory interface
we are heavily using the feign-annotation-error-decoder from OpenFeign in our project. Currently, We need to provide in each OpenFeign interface an additional configuration which provides an ErrorDecoder bean by using the following definition
```java
@Bean
public ErrorDecoder errorDecoder() {
    return AnnotationErrorDecoder.builderFor(OpenFeignInterface.class).build();
}
```
where `OpenFeignInterface.class` is the Interface with all the OpenFeign annotations plus the `@ErrorHandling` annotations.
This config is nearly every time the same except the class which needs to be provided to the builder.

My idea is to have a `ErrorDecoderFactory` interface with one method which takes the interface class as an input and return the ErrorDecoder, similar to what is currently available for creating loggers.

```java
public interface ErrorDecoderFactory {
	/**
	 * Factory method to provide a {@link ErrorDecoder} for a given {@link Class}.
	 * @param type the {@link Class} for which a {@link ErrorDecoder} instance is to be created
	 * @return a {@link ErrorDecoder} instance
	 */
	ErrorDecoder create(Class<?> type);
}
```

so in that case there needs only one bean available implementing the `ErrorDecoderFactory` interface in the context which allows us to create a specific `ErrorDecoder`. That simplifies and reduces the amount of code which needs to be written for each interface.

The implementation of checking if a factory bean is available can be done inside the `FeignClientFactoryBean` in two ways:
- if such a bean is available use it and create the `ErrorDecoder`, if not use the default `ErrorDecoder` by not setting the ErrorDecoder in the `Feign.Builder`
- provide a default `FeignClientFactoryBean` implementation which returns the default `ErrorDecoder` and adding an additional Bean definition to `FeignClientsConfiguration` with a `@ConditionalOnMissingBean`

tried to extend `FeignClientFactoryBean` but this implementation is package protected so it is not that easy to extend

**Additional context**
---

BigMichi1:
this factory will be used if no ErrorDecoder was found. This factory allows to create a ErrorDecoder based on the type. This fixes #308

### DIFF
--- a/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientErrorDecoderTests.java
+++ b/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientErrorDecoderTests.java
@@ -61,9 +61,8 @@ public class FeignClientErrorDecoderTests {
 
 	@Test
 	public void overrideErrorDecoder() {
-		ErrorDecoder.Default.class
-				.cast(this.context.getInstance("foo", ErrorDecoder.class));
-		ErrorDecoderImpl.class.cast(this.context.getInstance("bar", ErrorDecoder.class));
+		assertThat(this.context.getInstance("foo", ErrorDecoder.class)).isInstanceOf(ErrorDecoder.Default.class);
+		assertThat(this.context.getInstance("bar", ErrorDecoder.class)).isInstanceOf(ErrorDecoderImpl.class);
 	}
 
 	@Configuration(proxyBeanMethods = false)
--- a/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientErrorDecoderTests.java
+++ b/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientErrorDecoderTests.java
@@ -61,8 +61,10 @@ public class FeignClientErrorDecoderTests {
 
 	@Test
 	public void overrideErrorDecoder() {
-		assertThat(this.context.getInstance("foo", ErrorDecoder.class)).isInstanceOf(ErrorDecoder.Default.class);
-		assertThat(this.context.getInstance("bar", ErrorDecoder.class)).isInstanceOf(ErrorDecoderImpl.class);
+		assertThat(this.context.getInstance("foo", ErrorDecoder.class))
+				.isInstanceOf(ErrorDecoder.Default.class);
+		assertThat(this.context.getInstance("bar", ErrorDecoder.class))
+				.isNull();
 	}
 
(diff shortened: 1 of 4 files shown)


OlgaMaciaszek:
Hi @BigMichi1 Thanks for the PR. Will review tomorrow.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientErrorDecoderTests.java
```diff
@@ -0,0 +1,143 @@
...
+	private BarClient bar;
+
+	@Test
+	public void clientsAvailable() {
+		assertThat(this.foo).isNotNull();
+		assertThat(this.bar).isNotNull();
+	}
+
+	@Test
+	public void overrideErrorDecoder() {
```
OlgaMaciaszek: Please convert to a more readable test with actual assertions.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignErrorDecoderFactoryTests.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Michael Cramer
+ */
+public class FeignErrorDecoderFactoryTests {
+
```
OlgaMaciaszek: Please add a test that verifies that if `ErrorDecoder` is present it will be used even if a `FeignErrorDecoderFactory` is present.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientErrorDecoderTests.java
BigMichi1: done

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignErrorDecoderFactoryTests.java
BigMichi1: added a test for that

OlgaMaciaszek:
Merged within https://github.com/spring-cloud/spring-cloud-openfeign/commit/56bee946d8f7059661a6815d015d4a2c940a9b90.

PR #258, opened 2019-12-06
Fix GH-201: Multiple @RequestPart not Working

aaronjwhiteside:
Summary of changes follows:

- Delegate ALL requestBody encoding where `Content-Type` is `multipart/form-data` to the `SpringFormEncoder`.
- Introduce `RequestPartParameterProcessor` to deal with `@RequestPart` annotations, adds parameters to `MethodMetadata.formParams()`.
- Wrap `HttpMessageConversionException` in `EncodeException`.
- Add tests to verify expected behaviour.

However there still exists a gap in functionality where any user defined pojo will be serialized as a `Map<String,String>`, as there is currently no way for `SpringFormEncoder` to know about the `Content-Type` of the individual parts beyond `MultipartFile`, boxed primitive types (which are treated as `text/plain`) and other built-in types inherited from `FormEncoder`.

PR #314, opened 2020-03-29
Allow @RequestPart user-defined POJOs

darrenfoong:
I'm using Spring Cloud OpenFeign, and I have a use case where I need to send *multiple* files *and* user-defined POJOs as `multipart/form-data`, for example:

```
@PostMapping("/items")
SomeResponse uploadItems(@RequestPart List<SomePojo> items, @RequestPart List<MultipartFile> itemImages);
```

There was a recent fix in #258 that allowed the use of multiple `@RequestPart`s, but it was only for `MultipartFile` and boxed primitive types. This PR builds on the abovementioned PR by:

- Updating `SpringEncoder` to (optionally) use a new `SpringPojoFormEncoder` that comes with a `PojoFormWriter`, which serializes user-defined POJOs.
- `PojoFormWriter` is abstract; developers can implement missing methods to serialize POJOs to any string format. In this PR, I implemented `JsonPojoFormWriter`, which uses Jackson.
- Update `FeignClientsConfiguration` to optionally autowire a `PojoFormWriter` that will be used with `SpringEncoder` upon application start-up.
- Update `SpringMvcContract.processAnnotationsOnParameter` to **not** expand parameters when the `Content-Type` is `multipart/form-data`. This is because a simple, single-field POJO like `FeignClientTests.Hello` would have been converted to a `String`.
- Add unit tests.

I am not sure how best to fix the issue in `processAnnotationsOnParameter` and I don't have a good understand of what it does exactly. I'd like to know a better way to prevent expansion of simple POJOs like `FeignClientTests.Hello`.

I had earlier created a pull request on feign-form (OpenFeign/feign-form#89), but considering that the project is not regularly maintained, I decided to make a PR here instead.

I had also noticed that the recommended usage on feign-form's documentation is to create a `new SpringFormEncoder(new SpringEncoder(...))`, but because `SpringEncoder` itself has an *internal* instance of `SpringFormEncoder`, I thought it would be cleaner to make a PR here.


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -54,16 +54,24 @@
...
+	private SpringFormEncoder springFormEncoder;
 
 	private final ObjectFactory<HttpMessageConverters> messageConverters;
 
 	public SpringEncoder(ObjectFactory<HttpMessageConverters> messageConverters) {
+		this.springFormEncoder = new SpringFormEncoder();
+		this.messageConverters = messageConverters;
+	}
+
+	public SpringEncoder(PojoFormWriter writer,
```
spencergibb, line 72: Let's just autowire a `SpringFormEncoder`. we should be able to use `@Conditional*` annotations to create a special one with the new functionality.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/PojoFormWriter.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.web.multipart.MultipartFile;
+
+import static feign.form.ContentProcessor.CRLF;
+import static feign.form.util.PojoUtil.isUserPojo;
+
+/**
+ * @author Darren Foong
+ */
+public abstract class PojoFormWriter extends AbstractWriter {
```
spencergibb, line 37: Spring always prefixes abstract classes with `Abstract`. Pojo doesn't help naming here so maybe `AbstractFormWriter`?

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/PojoFormWriter.java
```diff
@@ -0,0 +1,83 @@
...
+			output.write(string);
+		}
+		catch (IOException e) {
+			throw new EncodeException(e.getMessage());
+		}
+	}
+
+	protected abstract MediaType getContentType();
+
+	protected abstract String writePojoAsString(Object object) throws IOException;
```
spencergibb, line 64: remove pojo here as well

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringPojoFormEncoder.java
```diff
@@ -0,0 +1,37 @@
...
+
+import feign.form.MultipartFormContentProcessor;
+import feign.form.spring.SpringFormEncoder;
+
+import static feign.form.ContentType.MULTIPART;
+
+/**
+ * @author Darren Foong
+ */
+public class SpringPojoFormEncoder extends SpringFormEncoder {
```
spencergibb, line 27: I don't think this needs to be a public class. Make it either private in `FeignClientsConfiguration` or just do this in a method in `FeignClientsConfiguration`. Add this there as well.
```
	@Autowired(required = false)
	private AbstractFormWriter abstractFormWriter;
```

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
darrenfoong: Something like this?

```
@Bean
@ConditionalOnBean(AbstractFormWriter.class)
private SpringFormEncoder springPojoFormEncoder() {
    return new SpringPojoFormEncoder(this.abstractFormWriter);
}

@Bean
@ConditionalOnMissingBean(AbstractFormWriter.class)
private SpringFormEncoder springFormEncoder() {
    return new SpringFormEncoder();
}
```

However, there will be two `Encoder` beans (`SpringEncoder` and `SpringFormEncoder`), which will cause problems at runtime. Any suggestions?

spencergibb: do it right in `FeignClientsConfiguration.feignEncoder(ObjectProvider<AbstractFormWriter> formWriter)`.

`ObjectProvider` has methods to help you decide. which constructor to call. Maybe it shouldn't be in the constructor, but a setter if it is truly optional.

darrenfoong: I've updated the code to use `ObjectProvider`, but it seems to be not very different from before. I use `getIfAvailable()` to get a `AbstractFormWriter` bean, but I still have to do a null check to choose the correct constructor for `SpringEncoder`. Previously, I null-checked the `@Autowired AbstractFormWriter`, which is consistent with the rest of the code in `FeignClientsConfiguration`.

I think the tricky part is that `SpringEncoder` *has* an instance of `SpringFormEncoder`. I've already tidied the constructors for `SpringEncoder` to now accept a `SpringFormEncoder` parameter, but I can't autowire `SpringFormEncoder` because it can't be a bean (see my previous comment). Is there an elegant way around this?

Thank you for your comments so far!

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -73,6 +80,9 @@
 	@Autowired(required = false)
 	private SpringDataWebProperties springDataWebProperties;
 
+	@Autowired(required = false)
+	private AbstractFormWriter formWriter;
```
OlgaMaciaszek, line 84: This is never used. Please remove.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -83,16 +93,36 @@ public Decoder feignDecoder() {
...
 
 	@Bean
 	@ConditionalOnClass(name = "org.springframework.data.domain.Pageable")
 	@ConditionalOnMissingBean
-	public Encoder feignEncoderPageable() {
-		PageableSpringEncoder encoder = new PageableSpringEncoder(
-				new SpringEncoder(this.messageConverters));
+	public Encoder feignEncoderPageable(
+			ObjectProvider<AbstractFormWriter> formWriterProvider) {
+		AbstractFormWriter formWriter = formWriterProvider.getIfAvailable();
```
OlgaMaciaszek, line 113: There's a very similar code section in the method above. Maybe it would be better to extract a method that takes `ObjectProvider<AbstractFormWriter> formWriterProvider` as argument and returns a correct `SpringEncoder` instance and reuse it while creating both beans?

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -54,16 +54,19 @@
...
  * @author Ahmad Mozafarnia
  * @author Aaron Whiteside
+ * @author Darren Foong
  */
 public class SpringEncoder implements Encoder {
 
 	private static final Log log = LogFactory.getLog(SpringEncoder.class);
 
-	private final SpringFormEncoder springFormEncoder = new SpringFormEncoder();
+	private SpringFormEncoder springFormEncoder;
```
OlgaMaciaszek, line 63: It's being always initialised in constructor - there's no need to remove the `final` keyword.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
```diff
@@ -54,16 +54,19 @@
...
 
 	private final ObjectFactory<HttpMessageConverters> messageConverters;
 
-	public SpringEncoder(ObjectFactory<HttpMessageConverters> messageConverters) {
+	public SpringEncoder(SpringFormEncoder springFormEncoder,
```
OlgaMaciaszek, line 67: Due to backwards compatibility issues, we cannot modify a `public` constructor in this way in a non-major release. This change can go within 2020-1 release train (you would have to submit your PR against the `master` branch then). Alternatively, you can provide an implementation with an additional constructor, while keeping the existing one as it was.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/AbstractFormWriter.java
```diff
@@ -0,0 +1,83 @@
...
+			Iterator<?> iterator = iterable.iterator();
+
+			return iterator.hasNext() && isType.test(iterator.next());
+		}
+		else {
+			return isType.test(object);
+		}
+	}
+
+}
```
OlgaMaciaszek, line 83: LGTM

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
```diff
@@ -272,6 +276,17 @@ private void checkOne(Method method, Object[] values, String fieldName) {
 	@Override
 	protected boolean processAnnotationsOnParameter(MethodMetadata data,
 			Annotation[] annotations, int paramIndex) {
+		boolean isMultipartFormData = false;
```
OlgaMaciaszek, line 279: Extract to a method.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringEncoder.java
darrenfoong: I went with the approach of providing two constructors for backward compatibility, but `FeignClientsConfiguration` is using the new constructor so the transition next time will involve fewer changes.

darrenfoong:
Thanks @spencergibb and @OlgaMaciaszek for reviewing and merging this PR!

PR #326, opened 2020-04-28
Ignore parent config option 2.2.x

matt62king:
Added ```FeignClientConfigurer``` with option to ignore parent configuration when giving a configuration to ```@FeignClient```

Fixes gh-132


OlgaMaciaszek:
Thanks a lot, @matt62king - will review this on Monday.

PR #347, opened 2020-05-17
update doc to clarify  the default retrying behavior

chang-chao:
Spring Cloud Feign will disable retrying , which is different from the Feign default one and better to be in the documentation.

https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java#L132-L134


OlgaMaciaszek:
@chang-chao That's a good change - thanks. Merged into `2.2.x` within https://github.com/spring-cloud/spring-cloud-openfeign/commit/84cf97b391e510eb0d24cd5d5d41d193c9f4f5c6 .

PR #352, opened 2020-06-12
Allow empty list for @RequestPart List<MultipartFile>

darrenfoong:
I have a use case where I need to send a list of `MultipartFile`s, but sometimes this list could be empty. Currently, to successfully send an empty list (i.e. the server receives an empty list as a method argument), I need to send `null` instead, which requires me to do a prior `isEmpty()` check.

(It is strange that sending `null` from the client results in an empty list received by the server, but this is documented [here](https://github.com/spring-projects/spring-framework/issues/20876#issuecomment-453465529).)

Sending an empty list now throws an exception:

```
feign.codec.EncodeException: class java.util.Collections$EmptyList is not a type supported by this encoder.
	at feign.codec.Encoder$Default.encode(Encoder.java:94)
	at feign.form.multipart.DelegateWriter.write(DelegateWriter.java:49)
	at feign.form.multipart.AbstractWriter.write(AbstractWriter.java:36)
	at feign.form.MultipartFormContentProcessor.process(MultipartFormContentProcessor.java:87)
	at feign.form.FormEncoder.encode(FormEncoder.java:105)
	at feign.form.spring.SpringFormEncoder.encode(SpringFormEncoder.java:84)
...
```

This PR adds an `EmptyCollectionWriter` to `MultipartFormContentProcessor` in the `SpringFormEncoder`. This `Writer` is a no-op.

It would be ideal to add the `EmptyCollectionWriter` in the [feign-form](https://github.com/OpenFeign/feign-form/) project instead, but that project is relatively inactive.

I also added a new test client endpoint that counts the number of files sent, to make sure that empty lists are not received as `null`, in case Spring Framework changes behaviour in the future.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.openfeign;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
 import com.fasterxml.jackson.databind.Module;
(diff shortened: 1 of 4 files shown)


OlgaMaciaszek:
Hi, @darrenfoong - thanks for the PR. Could you first check with the maintainers of feign-form if it's possible to add the `EmptyCollectionWriter`  there instead?

darrenfoong:
I've just created a PR for feign-form: https://github.com/OpenFeign/feign-form/pull/99. I hope the maintainers become active...

OlgaMaciaszek:
Closing as external issue.

PR #412, opened 2020-10-02
Add support for passing hints to LoadBalancer.

OlgaMaciaszek:


spencergibb (approved):

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/FeignBlockingLoadBalancerClient.java
```diff
@@ -76,4 +86,10 @@ public Client getDelegate() {
 		return delegate;
 	}
 
+	private String getHint(String serviceId) {
+		String defaultHint = properties.getHint().getOrDefault("default", "default");
```
ryanjbaxter, line 90: Seems like this could go in `LoadBalancerProperties`, then it would avoid the duplication in the reactive implementation

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/FeignBlockingLoadBalancerClient.java
```diff
@@ -76,4 +86,10 @@ public Client getDelegate() {
...
+		String hintPropertyValue = properties.getHint().get(serviceId);
```
ryanjbaxter, line 91: why not use `getOrDefault` here as well

ryanjbaxter (approved):

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/FeignBlockingLoadBalancerClient.java
OlgaMaciaszek: I'm using it, but then There are 3 options:  a `hint` for `serviceId`, a hint under key `default` or literally the `default` `String`, resolved in this order.

OlgaMaciaszek:
Fixes gh-411.

PR #418, opened 2020-10-23
Support loadbalancer lifecycle

OlgaMaciaszek:


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/RetryableFeignBlockingLoadBalancerClient.java
```diff
@@ -143,6 +169,28 @@ protected Response createResponse(Response response, URI uri) {
 		});
 	}
 
+	private Response executeWithLoadBalancerLifecycleProcessing(Request.Options options, Request feignRequest,
```
ryanjbaxter, line 172: Could this logic be shared between the clients?

ryanjbaxter (approved):

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/RetryableFeignBlockingLoadBalancerClient.java
OlgaMaciaszek: Will move it to a utility class.

PR #428, opened 2020-11-15
Fix for Path variables with / are not url encoded

### LINKED ISSUE #3: Path variables with / are not url encoded
With Feign directives the decodeSlash parameter of RequestLine directive deals with it, 
see https://stackoverflow.com/questions/43889140/feign-url-encode-path-params
There is no such thing with spring XXXMapping directive (which can not be modified).
(PR for fix to come)

minerasoft:
Fixes #3 

Feign clients do not encode slash `/` characters by default. To change this behaviour, set the `decodeSlash` property to `false`.

feign.client.decodeSlash=false

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
@@ -46,6 +46,12 @@ public class FeignClientProperties {
 
 	private Map<String, FeignClientConfiguration> config = new HashMap<>();
 
+	/**
+	 * Feign clients do not encode slash `/` characters by default. To change this
+	 * behavior, set the `decodeSlash` to `false`.
+	 */
+	private boolean decodeSlash = true;
+
 	public boolean isDefaultToProperties() {
 		return this.defaultToProperties;
 	}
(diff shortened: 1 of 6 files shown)


OlgaMaciaszek (changes requested):
@minerasoft This PR looks good. It's great you've added a lot of tests. However, since this is not a breaking change, we could include it already in the `Hoxton` release train, so could you submit your changes against the `2.2.x` branch instead of `master`?

OlgaMaciaszek:
@minerasoft  Since everything is fine with this PR, I'm going to cherry-pick your commits directly to `2.2.x` and also add some docs in order to get this change included in this weeks M6 release.

OlgaMaciaszek:
Merged within https://github.com/spring-cloud/spring-cloud-openfeign/commit/d419826ca0d17c94ffd8829442096c1d7fbf90b9.

minerasoft:
@OlgaMaciaszek Thanks you very much!

chErdemontas:
Hİ @minerasoft decode-slash=false effects all the clients in the codebase. What would you suggest if we have many clients and we only wants this config affect onyl 1 of them ?

OlgaMaciaszek:
Hello @chErdemontas. This is not currently supported and would be a separate enhancement. You may open a new issue to have it implemented, however, since the project is currently in maintenance only mode, the Spring Cloud team will not be working on it, so the only way it would get in is if you or someone else from the community submits a PR for it.

PR #430, opened 2020-11-16
Fix #205 - Move Spring Data Jackson Modules from FeignClientsConfiguration to FeignAutoConfiguration

### LINKED ISSUE #205: Unable to decode Page<T> response
> Spring Boot 2.1.6
> Spring Cloud OpenFeign 2.1.2
> Java 11

Hi there!

Looking at the Spring Cloud OpenFeign source code looks like it is able to decode Page responses, but something isn't working as expected. I have the following client:

```java
@RequestMapping(method = RequestMethod.POST, value = "users/find")
Page<UserResponse> findUsers(List<Long> usersId);
```

The `users/find` endpoint returns a `org.springframework.data.domain.Page` response, which should be decoded to `Page<UserResponse>`, but instead I get this exception:

```
Type definition error: [simple type, class org.springframework.data.domain.Page];
nested exception is com.fasterxml.jackson.databind.exc.InvalidDefinitionException:
Cannot construct instance of `org.springframework.data.domain.Page`
```

But shouldn't this project handle this through [PageJacksonModule.java](https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/PageJacksonModule.java) and [FeignClientsConfiguration.java](https://github.com/spring-cloud/spring-cloud-openfeign/blob/master/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java)?

The FeignClientsConfiguration is beaning the `return new PageJacksonModule()` class during application startup, and `PageJacksonModule` has the implementation that is needed to decode a response to `Page` through:

```java
delegate = new PageImpl<>(content, PageRequest.of(number, size),
    totalElements);
```

I didn't find any instructions in the documentation regarding the Page desserialization. Am I missing something?

Best regards,
Rafael Pacheco.

nkonev:
Fix https://github.com/spring-cloud/spring-cloud-openfeign/issues/205

@OlgaMaciaszek Could you check this PR please ?

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
@@ -24,6 +24,7 @@ import java.util.concurrent.TimeUnit;
 
 import javax.annotation.PreDestroy;
 
+import com.fasterxml.jackson.databind.Module;
 import feign.Client;
 import feign.Feign;
 import feign.httpclient.ApacheHttpClient;
@@ -48,6 +49,8 @@ import org.springframework.cloud.commons.httpclient.OkHttpClientConnectionPoolFa
 import org.springframework.cloud.commons.httpclient.OkHttpClientFactory;
 import org.springframework.cloud.openfeign.support.DefaultGzipDecoderConfiguration;
 import org.springframework.cloud.openfeign.support.FeignHttpClientProperties;
+import org.springframework.cloud.openfeign.support.PageJacksonModule;
+import org.springframework.cloud.openfeign.support.SortJacksonModule;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
@@ -55,6 +58,8 @@ import org.springframework.context.annotation.Import;
 /**
  * @author Spencer Gibb
  * @author Julien Roy
+ * @author Grzegorz Poznachowski
+ * @author Nikita Konev
  */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnClass(Feign.class)
@@ -76,6 +81,22 @@ public class FeignAutoConfiguration {
 		FeignContext context = new FeignContext();
 		context.setConfigurations(this.configurations);
 		return context;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(PageJacksonModule.class)
+	@ConditionalOnClass(name = "org.springframework.data.domain.Page")
+	@ConditionalOnProperty(value = "feign.autoconfiguration.jackson.enabled", havingValue = "true")
+	public Module pageJacksonModule() {
+		return new PageJacksonModule();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(SortJacksonModule.class)
+	@ConditionalOnClass(name = "org.springframework.data.domain.Sort")
+	@ConditionalOnProperty(value = "feign.autoconfiguration.jackson.enabled", havingValue = "true")
+	public Module sortModule() {
+		return new SortJacksonModule();
 	}
 
 	@Configuration(proxyBeanMethods = false)
--- a/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
@@ -663,6 +663,14 @@ To work around this problem you can use an `ObjectProvider` when autowiring your
 ObjectProvider<TestFeginClient> testFeginClient;
 ----
 
+=== Spring Data Support
+You may consider enabling Jackson Modules for the support `org.springframework.data.domain.Page` and `org.springframework.data.domain.Sort` decoding.
+
+[source,java]
+----
+feign.autoconfiguration.jackson.enabled=true
+----
+
 == Configuration properties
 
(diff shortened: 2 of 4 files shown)


nkonev:
Any chance to be reviewed ?

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -78,6 +83,18 @@ public FeignContext feignContext() {
 		return context;
 	}
 
+	@Bean
```
OlgaMaciaszek, line 86: Let's add `@ConditionalOnMissingBean` as well.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -78,6 +83,18 @@ public FeignContext feignContext() {
...
+	@ConditionalOnClass(name = "org.springframework.data.domain.Page")
+	public Module pageJacksonModule() {
+		return new PageJacksonModule();
+	}
+
+	@Bean
```
OlgaMaciaszek, line 92: Let's add `@ConditionalOnMissingBean` as well.

OlgaMaciaszek (changes requested):
Hi, @nkonev Thanks for the PR. Have added some comments. Please have a look.

nkonev:
@OlgaMaciaszek Hi, thank you, I've done.

nkonev:
@OlgaMaciaszek Hello! Do I have any chance to be reviewed/merged ?
This change looks simple, it isn't a big refactoring.

OlgaMaciaszek:
Hi, @nkonev Sorry, I was away. Let me have a look at it now.

OlgaMaciaszek:
@nkonev Have done some tests on it. This does solve the issue. However, I've realised the conditional will not be enough, as users might also register their modules just with with `Module` type also we could possibly have bean naming clashes. In order to avoid any backwards compatibility issue, let's make it opt-in by adding a new property, for example, `feign.autoconfiguration.jackson.enabled` and add a `@ConditionalOnProperty` `havingValue` `true`.

nkonev:
@OlgaMaciaszek  Thank you for reply.
Backward compatibility is pain.
I think to add it with TODO to change it in the future Spring Clod 2021 release.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -78,6 +83,22 @@ public FeignContext feignContext() {
 		return context;
 	}
 
+	@Bean
+	@ConditionalOnMissingBean(PageJacksonModule.class)
+	@ConditionalOnClass(name = "org.springframework.data.domain.Page")
+	@ConditionalOnProperty(value = "feign.autoconfiguration.jackson.enabled", havingValue = "true", matchIfMissing = true)
```
OlgaMaciaszek, line 89: Since we want it to be opt-in, it should be `matchIfMissing` = `false`, which probably is the default.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -78,6 +83,22 @@ public FeignContext feignContext() {
...
+	public Module pageJacksonModule() {
+		return new PageJacksonModule();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(SortJacksonModule.class)
+	@ConditionalOnClass(name = "org.springframework.data.domain.Sort")
+	@ConditionalOnProperty(value = "feign.autoconfiguration.jackson.enabled", havingValue = "true", matchIfMissing = true)
```
OlgaMaciaszek, line 97: Since we want it to be opt-in, it should be `matchIfMissing` = `false`, which probably is the default.

OlgaMaciaszek (changes requested):
Have added comments. Please add changes. Also, please add documentation in `spring-cloud-openfeign.adoc` describing the feature and the new property.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
nkonev: So that way we will preserve old behavior, and in order to fix bug user should manually set feign.autoconfiguration.jackson.enabled=true. Is it real need to preserve old buggy behavior ?

nkonev: Or may be it has the meaning to do matchIfMissing=false with TODO Change in Spring Cloud 2021 ?

OlgaMaciaszek (approved):

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
OlgaMaciaszek: We could do it in a major. Unfortunately, a major has been just released, so we need to wait for the next one.

PR #445, opened 2020-12-14
Moved the code from Spring Cloud Security

### LINKED ISSUE #444: Moving spring-cloud-security code
related - https://github.com/spring-cloud/spring-cloud-security/issues/231

marcingrzejszczak:
- moved the interceptor
- added configuration for the interceptor

Maybe @dsyer or @spencergibb will remember why there was no auto configuration for the oauth2 interceptor for Feign in the first place? I couldn't find one in Spring Cloud Security.

fixes #444

### DIFF
--- a/spring-cloud-openfeign-core/pom.xml
+++ b/spring-cloud-openfeign-core/pom.xml
@@ -120,6 +120,11 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.security.oauth.boot</groupId>
+			<artifactId>spring-security-oauth2-autoconfigure</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-autoconfigure-processor</artifactId>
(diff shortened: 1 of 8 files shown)


spring-cloud-openfeign-core/pom.xml
```diff
@@ -119,6 +119,11 @@
 			<artifactId>okhttp</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.security.oauth.boot</groupId>
+			<artifactId>spring-security-oauth2-autoconfigure</artifactId>
```
spencergibb, line 124: Was hoping to move to the new stuff, but just having this moved is better. I think there's an open issue, that can come in a SR

spencergibb (approved):

spring-cloud-openfeign-core/pom.xml
mteodori: this one that should be moved to this project probably spring-cloud/spring-cloud-security#173
in the comments there are links to a couple of possible implementation using Spring Security only

Issue #279, opened 2020-01-18
Add support for spring-cloud-circuitbreaker


spencergibb:
Is this possible?

ryanjbaxter:
See https://github.com/spring-cloud/spring-cloud-circuitbreaker/issues/3 as well

PR #446, opened 2020-12-15
Added support for Spring Cloud CircuitBreaker

marcingrzejszczak:
fixes gh-279

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2013-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.openfeign;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import feign.InvocationHandlerFactory;
+import feign.Target;
+
+import org.springframework.cloud.client.circuitbreaker.CircuitBreaker;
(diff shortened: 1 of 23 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+				return false;
+			}
+		}
+		else if ("hashCode".equals(method.getName())) {
+			return hashCode();
+		}
+		else if ("toString".equals(method.getName())) {
+			return toString();
+		}
+		String circuitName = this.feignClientName + "_" + method.getName();
```
marcingrzejszczak, line 78: Is this a good name?

spring-cloud-openfeign-dependencies/pom.xml
```diff
@@ -50,11 +50,6 @@
 				<artifactId>feign-httpclient</artifactId>
 				<version>${feign.version}</version>
 			</dependency>
-			<dependency>
```
marcingrzejszczak, line 53: This is a breaking change!

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java
OlgaMaciaszek: Seems ok to me. @ryanjbaxter wdyt?

spring-cloud-openfeign-dependencies/pom.xml
OlgaMaciaszek: Let's keep it in `master` but make sure we don't have it when we add these changes to `2.2.x`.

OlgaMaciaszek (approved):
Looks good to me. One thing to keep in mind is that these changes should be added to Hoxton (minus the breaking one) in an opt-in fashion (Hystrix used by default, possibility to switch to using CB).

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java
ryanjbaxter: Documenting the name will be important.  Users need to know the circuit breaker name in order to customize it if they want https://docs.spring.io/spring-cloud-circuitbreaker/docs/1.0.4.RELEASE/reference/html/#specific-circuit-breaker-configuration

ryanjbaxter (approved):

ryanjbaxter:
Back porting this to Hoxton is going to be messy, but doable I suppose

marcingrzejszczak:
Closed via https://github.com/spring-cloud/spring-cloud-openfeign/commit/63a524843d3cfb9c6c1de45ca6cff5e1ea6e6a26

marcingrzejszczak:
@ryanjbaxter https://github.com/spring-cloud/spring-cloud-openfeign/commit/63a524843d3cfb9c6c1de45ca6cff5e1ea6e6a26#diff-b258ace955d38ee849bf3d4a8e0afdb84712346131522e39bb4782b1cffc8506R332 - added such text

PR #447, opened 2020-12-18
Support LB stats.

OlgaMaciaszek:


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/LoadBalancerUtils.java
```diff
@@ -84,8 +86,9 @@ static RequestData buildRequestData(Request request) {
 
 	static Response executeWithLoadBalancerLifecycleProcessing(Client feignClient, Request.Options options,
 			Request feignRequest, org.springframework.cloud.client.loadbalancer.Response<ServiceInstance> lbResponse,
+			org.springframework.cloud.client.loadbalancer.Request lbRequest,
 			Set<LoadBalancerLifecycle> supportedLifecycleProcessors) throws IOException {
-		return executeWithLoadBalancerLifecycleProcessing(feignClient, options, feignRequest, lbResponse,
+		return executeWithLoadBalancerLifecycleProcessing(feignClient, options, feignRequest, lbResponse, lbRequest,
```
spencergibb, line 91: This was the thing that prompted my question about parameter order since, lbrequest should come before lbresponse.

spencergibb (approved):
LGTM

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/LoadBalancerUtils.java
OlgaMaciaszek: ok, will switch it.

PR #451, opened 2020-12-27
Support to set loadBalancerKey from request header

alan-tang-tt:
When I write myself `MyRule` implements `IRule`, the key is always null, like below:

```java
public class MyRule extends AbstractLoadBalancerRule {

    @Override
    public Server choose(Object key) {
          // here the key is always null
    }
}
```

By watching the source code, I find there's no where to set this `loadBalancerKey`. So I push this PR to resolve this problem by overriding `FeignLoadBalancer#customizeLoadBalancerCommandBuilder` method.

```java
         // override FeignLoadBalancer#customizeLoadBalancerCommandBuilder
	@Override
	protected void customizeLoadBalancerCommandBuilder(RibbonRequest request, IClientConfig config, LoadBalancerCommand.Builder<RibbonResponse> builder) {
		builder.withServerLocator(request.request.headers().get("loadBalancerKey"));
	}
```

By this way, I can put my `loadBalancerKey` from `feign.RequestInterceptor` implementation like this:

```java
public class MyRequestInterceptor implements RequestInterceptor {
    @Override
    public void apply(RequestTemplate template) {
        template.header("loadBalancerKey", "gray");
    }
}
```

Then in `MyRule` I can get the key:

```java
public class MyRule extends AbstractLoadBalancerRule {

    @Override
    public Server choose(Object key) {
          // no the key equals to gray
    }
}
```

This is all.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/FeignLoadBalancer.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/FeignLoadBalancer.java
@@ -33,6 +33,7 @@ import com.netflix.client.RetryHandler;
 import com.netflix.client.config.IClientConfig;
 import com.netflix.loadbalancer.ILoadBalancer;
 import com.netflix.loadbalancer.Server;
+import com.netflix.loadbalancer.reactive.LoadBalancerCommand;
 import feign.Client;
 import feign.Request;
 import feign.Response;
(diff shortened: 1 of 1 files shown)


OlgaMaciaszek (changes requested):
@alan-tang-tt Thanks for the PR. The idea is good, however some more work has to be done on the PR.
Please verify if the headers contain the header you want to use.
Please change the default header name to `X-LodBalancer-Key`.
Please allow changing header name via properties.
Please add tests.
Please add documentation.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/FeignLoadBalancer.java
```diff
@@ -93,6 +94,11 @@ public RibbonResponse execute(RibbonRequest request, IClientConfig configOverrid
 		return new RibbonResponse(request.getUri(), response);
 	}
 
+	@Override
+	protected void customizeLoadBalancerCommandBuilder(RibbonRequest request, IClientConfig config, LoadBalancerCommand.Builder<RibbonResponse> builder) {
+		builder.withServerLocator(request.request.headers().get("loadBalancerKey"));
```
kimmking, line 99: And consider other parameter like url parameter？

OlgaMaciaszek:
Closing due to lack of activity.

Issue #441, opened 2020-12-09
FeignClient gets created too early for contract tests


akutschera:
A while ago I created a bug in Spring Cloud Contract (https://github.com/spring-cloud/spring-cloud-contract/issues/1303)
After some more debugging it looks like the FeignRegistrar registers the feign client *before* the StubRunner is created, therefore the random port of the stub runner is not set for the feign client (the property that the stubrunner sets is "stubrunner.runningstubs.test-webapp.port", but it sets it too late).
In my example I would expect the test class annotations to be evaluated before the production annotations are.

See the Spring Cloud Contract issue for a sample demonstrating the unexpected behaviour.

spencergibb:
Feign Clients have timing issues. You can not use them in constructors or post construct. I'm not sure there's anything we can do

akutschera:
I don't understand your answer. I am autowiring a FeignClient in my test class. That FeignClient has an a url property that gets evaluated before the test class annotations are fully processed (that's why I cannot connect to the test wiremock).
It's all working fine when I use a static port for this.
It's not working when I want to use a random port set by the test. It would work if I used reflection to set the url of the created feignClient proxy which looks hacky to me.
What are the timing issues you are speaking of?

lukasz-gryzbon:
I also hit the same problem - twice.

To give a bit of a background, my team runs a number of projects. Each has Cucumber component tests that use Feign clients. So basically, we inject Feign clients into the compoent tests to call the service being tested. The services itself uses Feign clients to call other services, which in case of component tests, are Wiremock server(s) span up by the tests themselves.

The problems we have encountered that I feel are related to what was reported here:

1. Because the Feign clients are created so early in the process of spinning up the service (Spring Boot context), when we try to use a random port like this `@SpringBootTest(webEnvironment = RANDOM_PORT)` and inject the `${server.port}` to the properties file as part of the URL proeperty used for `@FeignClient(url = "${url}")`, the `server.port` value is not available at the time of creating the Feign client, so we can't use it. We had to fall back to a static port and make sure each project uses a different one so there is not port conflict when building multiple projects on a single machine.
[...]

Issue #443, opened 2020-12-14
Deadlock may occure when Openfeign works with deferred JPA repository


michaldo:
Spring Cloud version "Hoxton.SR9"

Under some condition deadlock happens with  Openfeign + deferred JPA repository. I can reproduce the case only in tests.

I created a simple project to reproduce the problem: https://github.com/michaldo/feign-jpa-deffered-deadlock

I think the same case is reported on https://stackoverflow.com/questions/65078382/spring-boot-context-load-test-hangs

I guess the reason is that `FeignClientFactoryBean`  does not extend `AbstractFactoryBean`

OlgaMaciaszek:
@michaldo Thanks for reporting this.

Further analysis:
From [here](https://github.com/spring-cloud/spring-cloud-openfeign/blob/5f932b01589370e933a4a14efe1d4464f340606f/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java#L104) new context is being created, which triggers `ContextRefreshedEvent`. `DataSourceInitializedPublisher` [listens on this event and tries to get `dataSourceInitialization` future](https://github.com/spring-projects/spring-boot/blob/bf9c67b793d67c1ffb569fd3c49114bacdb60755/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/DataSourceInitializedPublisher.java#L148). This is causing a deadlock. 
`dataSourceIniitialization` is being initialized within the `postProcessEntityManagerFactory` method.

OlgaMaciaszek:
The problem seems to be in Spring Boot. Have created https://github.com/spring-projects/spring-boot/issues/24966.  Also, upgrading to Boot `2.3.7` should help as the deferred mode was set to non-default (https://github.com/spring-projects/spring-boot/issues/24249).
Closing in favour of the Spring Boot issues.

michaldo:
Initially Spring Boot guys told it is OpenFeing issue :)

PR #455, opened 2021-01-12
Lazy openfeign bean registration

marcingrzejszczak:
without this change we're eagerly resolving placeholder properties in passed URLs / names. Since this happens at bean definition level it's extremely early e.g. Spring Cloud Contract has not yet registered any placeholders that Feign could try to consume.
with this change we're changing the bean to become lazy initialized and we resolve the placeholder values at runtime - as late as possible.

problems:
- one of the tests uses the factory bean - there's no more factory bean since the instance supplier for the bean registration acts as one

related issue https://github.com/spring-cloud/spring-cloud-contract/issues/1303

### DIFF
--- a/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientUsingConfigurerTest.java
+++ b/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientUsingConfigurerTest.java
@@ -22,6 +22,7 @@ import java.util.List;
 import feign.Feign;
 import feign.Logger;
 import feign.RequestInterceptor;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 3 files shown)


ryanjbaxter (approved):

OlgaMaciaszek (changes requested):
Looks good, however, I don't think it makes sense to ignore the test unless we have a plan in place to makes changes to it. We should either modify this one or create a new one or, if that's not possible, remove the existing one.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientUsingConfigurerTest.java
```diff
@@ -47,9 +48,10 @@
 		"feign.client.config.default.loggerLevel=full",
 		"feign.client.config.default.requestInterceptors[0]=org.springframework.cloud.openfeign.FeignClientUsingPropertiesTests.FooRequestInterceptor",
 		"feign.client.config.default.requestInterceptors[1]=org.springframework.cloud.openfeign.FeignClientUsingPropertiesTests.BarRequestInterceptor" })
+@Ignore("I don't know how to test this since the factory bean is no longer created")
```
spencergibb, line 51: Why is the factory bean no longer created?

lukasz-gryzbon: From the initial PR comment: `there's no more factory bean since the instance supplier for the bean registration acts as one`

spencergibb: Then that class should be removed and the test removed.

OlgaMaciaszek (commented):
LGTM

michaldo:
I would like you consider https://github.com/spring-cloud/spring-cloud-openfeign/issues/443

> I guess the reason is that FeignClientFactoryBean does not extend AbstractFactoryBean

Probably reusing AbstractFactoryBean patterns will decrease chance Spring-related problems

marcingrzejszczak:
But we no longer register `FeignClientFactoryBean` as a bean. Can you check @michaldo if with this change the problem in #443 didn't go away?

michaldo:
#443 still exists on 3.0.1-SNAPSHOT

It looks very suspicious to me that AbstractFactoryBean#getObject returns *"early singleton proxy"* and FeignClientFactoryBean#getObject returns *"concrete bean"*. There must be a reason that base Spring class returns proxy

OlgaMaciaszek:
Fixes gh-441.

Issue #456, opened 2021-01-12
ResponseEntity headers broken w/ spring-cloud 2020.0.0


o2-mesmer:
I'm trying to upgrade my system from spring-cloud Hoxton.SR8 + spring-boot 2.34.-RELEASE to spring-cloud-2020.0.0 + spring-boot 2.4.1.

Some of my tests started failing upon this upgrade, and investigation shows what seems to me a rather serious issue that I'd like to bring to your attention.

I have a simple echo service that takes a query param and echoes it back. I call it with a simple Feign client:
```
@FeignClient(name = "echo-service")
public interface EchoClient {
    @GetMapping(path = "/echo", produces = MediaType.APPLICATION_JSON_VALUE)
    ResponseEntity<Echo> jsonEcho(@RequestParam("m") String message);
}
```

My test checks the Content-Type header of the response:
```
        String message = "Hello JSON!";
        ResponseEntity<Echo> entity = client.jsonEcho(message);
        assertThat(entity.getHeaders().getContentType()).isEqualTo(MediaType.APPLICATION_JSON);
```

The test fails, because response headers seem to be broken in this latest version. I don't know whence the change stems, but it seems pretty clear it won't work properly:
- `feign.SynchronousMethodHandler` calls `feign.Response#build`
- `feign.Response` converts every header key *to lower case* in `feign.Response#caseInsensitiveCopyOf`
- Response (in my case) gets wrapped into a `org.springframework.http.ResponseEntity`
- *ResponseEntity's headers are case-sensitive*. The assertion in the test above fails because `entity.getHeaders().getContentType()` is `null`!

o2-mesmer:
The same test **passes** with spring-cloud Hoxton.SR9 and spring-boot 2.3.7.RELEASE, because `ResponseEntity#getHeaders` eventually queries a `org.springframework.util.LinkedCaseInsensitiveMap` (`spring-core-5.2.12.RELEASE` and `spring-web-5.2.12.RELEASE`), which handles header keys cases-insensitively.

martinsefcik:
Maybe `ResponseEntityDecoder` in `createResponse` method should use `HttpHeaders` (and it's default constructor which uses LinkedCaseInsensitiveMap) instead of case sensitive `LinkedMultiValueMap` to build headers for `ResponseEntity`.

martinsefcik:
[...]

PR #472, opened 2021-01-27
Use case-insensitive ResponseEntity headers. Fixes gh-456.

OlgaMaciaszek:


spencergibb (approved):
LGTM, assuming the tests fail without the fix

PR #479, opened 2021-02-08
Fixes #355 by importing the OpenFeign bom

### LINKED ISSUE #355: Import feign-bom instead of managing specific OpenFeign modules
Managing specific modules prevents safe inclusion of OpenFeign modules not used by Spring Cloud explicitly because version needs to be redeclared in application.

Extracted from #354

timpeeters:
Fixes #355 by importing the OpenFeign bom.


OlgaMaciaszek:
Fixes gh-355.

PR #483, opened 2021-02-16
Fixes #482 No bean found of type Targeter.

### LINKED ISSUE #482: No bean found of type Targeter after upgrade to Hoxton SR10
After upgrading from Hoxton SR9 to Hoxton SR10 we suddenly get the following error during startup:

`No bean found of type interface org.springframework.cloud.openfeign.Targeter`

I must add that we have always excluded `feign-hystrix` in the past because we did not use it and the dependency is rather "heavy". 

```
<dependency>
    <groupId>org.springframework.cloud</groupId>
    <artifactId>spring-cloud-starter-openfeign</artifactId>
    <exclusions>
        <exclusion>
            <artifactId>feign-hystrix</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

`FeignAutoConfiguration` supported this before Hoxton SR10 by conditionally loading the `DefaultTargeter` instead of the `HystrixTargeter`:

```
@Configuration(proxyBeanMethods = false)
@ConditionalOnMissingClass("feign.hystrix.HystrixFeign")
protected static class DefaultFeignTargeterConfiguration {
    @Bean
    @ConditionalOnMissingBean
    public Targeter feignTargeter() {
        return new DefaultTargeter();
    }
}
```

But as of Hoxton SR10, Spring Cloud Circuit Breaker support was added to `FeignAutoConfiguration` and the conditionals have changed. 

```
@Configuration(proxyBeanMethods = false)
@ConditionalOnMissingClass({ "feign.hystrix.HystrixFeign",
		"org.springframework.cloud.client.circuitbreaker.CircuitBreaker" })
protected static class DefaultFeignTargeterConfiguration {
    @Bean
    @ConditionalOnMissingBean
    public Targeter feignTargeter() {
        return new DefaultTargeter();
    }
}
```

Now the `DefaultFeignTargeterConfiguration` is only loaded if both `HystrixFeign` and `CircuitBreaker` interface are not on the classpath. The `CircuitBreaker` interface however is always included through `spring-cloud-commons` and thus the `DefaultFeignTargeterConfiguration` condition will never load.

Plus, the newly added `CircuitBreakerPresentFeignTargeterConfiguration` also tries to register a `DefaultTargeter` in case:

- `CircuitBreaker` is on the classpath (always true) 
- `feign.circuitbreaker.enabled` is set to true (which defaults to false).
- There is no `CircuitBreakerFactory` bean defined

This however feels strange as one would have to set `feign.circuitbreaker.enabled` to `true` without actually using circuit breaking just to get the `DefaultTargeter` to load.

```
@Configuration(proxyBeanMethods = false)
@ConditionalOnClass(CircuitBreaker.class)
@ConditionalOnProperty("feign.circuitbreaker.enabled")
protected static class CircuitBreakerPresentFeignTargeterConfiguration {
    @Bean
    @ConditionalOnMissingBean(CircuitBreakerFactory.class)
    public Targeter defaultFeignTargeter() {
        return new DefaultTargeter();
    }

    @Bean
    @ConditionalOnMissingBean
    @ConditionalOnBean(CircuitBreakerFactory.class)
    public Targeter circuitBreakerFeignTargeter(CircuitBreakerFactory circuitBreakerFactory) {
        return new FeignCircuitBreakerTargeter(circuitBreakerFactory);
    }
}
```

In its current state, I have the feeling that the `DefaultFeignTargeterConfiguration` is unreachable code and that it would be better to merge the 2 `DefaultTargeter` definitions into one `Configuration` that is conditionally loaded when both Hystrix or Spring Cloud Circuit Breaker are not explicitly enabled.

I'll be happy to submit a pull request if you can provide some feedback/guidance on this.

timpeeters:
Changes:

- Removed `@ConditionalOnMissingClass(HystrixFeign, CircuitBreaker)` on `DefaultFeignTargeterConfiguration`. Both classes are present by default and this causes the DefaultTargeter not to be considered even if Hystrix and CircuitBreaker are disabled.
- Position `DefaultTargeter` behind `HystrixTargeter` using `@AutoConfigureAfter`
- Position `HystrixTargeter` behind `CircuitBreakerTarger` using `@AutoConfigureAfter`
- Make it possible to disable `HystrixTargeter` by setting `feign.hystrix.enabled` to `false`. Keep default behaviour from previous release where  `HystrixTargeter` was the default (`matchIfMissing`).
- Only enable `CircuitBreakerTargeter` if `feign.circuitbreaker.enabled` is true

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
@@ -40,7 +40,7 @@ import org.apache.http.conn.HttpClientConnectionManager;
 import org.apache.http.impl.client.CloseableHttpClient;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -59,6 +59,7 @@ import org.springframework.cloud.openfeign.support.FeignHttpClientProperties;
 import org.springframework.cloud.openfeign.support.PageJacksonModule;
 import org.springframework.cloud.openfeign.support.SortJacksonModule;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.data.domain.Page;
@@ -116,7 +117,7 @@ public class FeignAutoConfiguration {
 	}
 
 	@Configuration(proxyBeanMethods = false)
-	@AutoConfigureAfter(HystrixFeignTargeterConfiguration.class)
+	@Conditional(DefaultFeignTargeterConditions.class)
 	protected static class DefaultFeignTargeterConfiguration {
 
 		@Bean
@@ -128,10 +129,10 @@ public class FeignAutoConfiguration {
 	}
 
 	@Configuration(proxyBeanMethods = false)
+	@Conditional(FeignCircuitBreakerDisabledConditions.class)
 	@ConditionalOnClass(name = "feign.hystrix.HystrixFeign")
 	@ConditionalOnProperty(value = "feign.hystrix.enabled", havingValue = "true",
 			matchIfMissing = true)
-	@AutoConfigureAfter(CircuitBreakerPresentFeignTargeterConfiguration.class)
 	protected static class HystrixFeignTargeterConfiguration {
 
 		@Bean
@@ -280,6 +281,24 @@ public class FeignAutoConfiguration {
 		@ConditionalOnMissingBean(Client.class)
 		public Client feignClient(okhttp3.OkHttpClient client) {
 			return new OkHttpClient(client);
+		}
+
+	}
+
+	static class DefaultFeignTargeterConditions extends AllNestedConditions {
+
+		DefaultFeignTargeterConditions() {
+			super(ConfigurationPhase.PARSE_CONFIGURATION);
+		}
+
+		@Conditional(FeignCircuitBreakerDisabledConditions.class)
+		static class FeignCircuitBreakerDisabled {
+
+		}
+
+		@Conditional(HystrixDisabledConditions.class)
+		static class HystrixDisabled {
+
 		}
 
 	}
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
@@ -40,6 +40,7 @@ import org.apache.http.conn.HttpClientConnectionManager;
 import org.apache.http.impl.client.CloseableHttpClient;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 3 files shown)


timpeeters:
@OlgaMaciaszek the build is referencing some missing snapshot dependencies: 
org.springframework.cloud:spring-cloud-netflix-dependencies:pom:2.2.8.BUILD-SNAPSHOT
org.springframework.cloud:spring-cloud-commons-dependencies:pom:2.2.8.BUILD-SNAPSHOT
(probably not related to the changes I've made?)

Let me know if you require any other changes. Given there are multiple @Conditional... options to get this working you might see things differently? 
I've also added some tests to get coverage on the changed code.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -114,8 +115,7 @@ public SortJacksonModule sortModule() {
 	}
 
 	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnMissingClass({ "feign.hystrix.HystrixFeign",
```
OlgaMaciaszek, line 117: Let's keep the onClass conditional  to handle any exclusions.

OlgaMaciaszek (changes requested):
@timpeeters -  in general looks good, however, have added one comment - please address. Also, please add your full name with `@author` tag to the javadoc.

OlgaMaciaszek:
Fixes gh-482.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
timpeeters: Author tags added.

I might be missing something but I'm afraid that when adding the `@ConditionalOnMissingClass` tag again, we are still missing the following cases:

1. User excludes `feign-hystrix` and does _not_ set `feign.circuitbreaker.enabled=true`. In this case the `@ConditionalOnMissingClass` will not load the `DefaultTargeter` because `CircuitBreaker` is still on the classpath and `CircuitBreakerTargeter` is not enabled.
2. Similarly, user excludes `feign-hystrix` and set `feign.circuitbreaker.enabled=true` but does not define a `CircuitBreakerFactory` bean.
3. User excludes `spring-cloud-commons` (unlikely because a lot will fail I guess), and disables Hystrix by setting `feign.hystrix.enabled=false`. In this case the `@ConditionalOnMissingClass` will again skip loading the `DefaultTargeter` because `HystrixFeign` is still on the classpath.

I therefor went for the approach using @AutoConfigureAfter to first:

1. Try to load `FeignCircuitBreakerTargeter`, if `CircuitBreaker` is on the classpath (usually the case I guess), `feign.circuitbreaker.enabled=true`, a `CircuitBreakerFactory` bean is present and the user did not define its own Feign `Targeter`.
2. If step 1 did not create a `FeignCircuitBreakerTargeter`, create a `HystrixTargeter` if `HystrixFeign` is on the classpath and `feign.hystrix.enabled` is not set to false.
3. In all other case, just create the `DefaultTargeter`.

Looking forward to your feedback.

OlgaMaciaszek: Yes, I should have specified more precisely - I meant creating a condition that checks that either the class is missing or the property is `false`. However, I agree that it might also work fine just by the config order as you suggest. If you would like to keep this approach, please add tests that will verify behaviour for exclusion (for example by using `ApplicationContextRunner` and `FilteredClassLoader`, but not necessarily - let's just make sure it's tested).

timpeeters: I see, I like the idea about the custom condition. I'll see if can make that work. Thanks for the guidance.

timpeeters: Apparently there already was a `FeignCircuitBreakerDisabledConditions` class that combined those checks. I created a new `HystrixDisabledConditions` class similar to that and combined both in one condition using `AllNestedConditions`. You were right, this looks better already.

I've also added tests using the `ApplicationContextRunner` and `FilterClassLoader` for all scenarios I could think of.

Maybe 2 questions that remain:

- I currently maintain the previous behaviour that a HystrixTargeter is created when the class is present even though it is not explicitly enabled using `feign.hystrix.enabled=true`. I would have preferred to return a DefaultTargeter in this case just like it is the case for the `FeignCircuitBreakerTargeter` but I guess keeping it backwards compatible is also important. I leave it up to you to decide.
- I opted to prefer the `FeignCircuitBreakerTargeter` over the `HystrixTargeter` when both are on the classpath and enabled via properties. This was implemented by adding the existing `FeignCircuitBreakerDisabledConditions` to the `HystrixFeignTargeterConfiguration`. Again this is up for debate.

OlgaMaciaszek: Yes, this looks better to me too. 
For the first question, we definitely should maintain backwards compatibility. Regarding the second one, this makes sense, cause `FeignCircuitBreakerTargeter` has to be explicitly enabled.

OlgaMaciaszek (approved):
LGTM.

PR #503, opened 2021-03-04
Propose a brand new contract interface `AsyncFeignClient`

thanhj:
Propose a brand new contract interface because the current one `FeignClient` is too tightly coupled with Non-Async classes and hard to enhance.


OlgaMaciaszek:
I'm not sure we need an entirely new annotation. It seems like an annotation parameter should be enough to indicate which Feign implementation to use.

thanhj:
The reason a new annotation is better because the Async one doesn't support current blocking operations such as Retry, LoadBalancer, Ribbon, etc. So it should not be mixed with the current annotation. 
I could continue to work on the rest of the implementation if I can get your agreement on this.

thanhj:
I try to make it a new property in the exiting annotation `@FeignClient`. Close the PR for now.

PR #518, opened 2021-03-30
Allow load balancer uri in uri attribute

### LINKED ISSUE #489: Support lb:// url in FeingClient annotation
# Description

In a project where is used spring cloud gateway, feign client (with target put in configuration), we ended up having feign clients declared as follows

```java
@FeignClient(name="${service.name:any-name}", url="${service.direct-url:}")
```

With configuration as follows

```yaml
service:
   name: service-id
   # url: http://direct-url (only used when service discovery is not available)
```

This way we are able, by configuration, to either run in service discovery/load balancing  mode and direct mode.

I kind of like the way spring cloud gateway handle that by allowing `lb://` protocol. Allowing this *protocol* in `url` attribute would allow the  following implementation

```java
@FeignClient(name="technical-name", url="${service.uri}")
```

with following configuration

```yaml
service:
   uri: "lb://service-id" # or "http//direct-url"
```

this way it will let common behaviour, allowing to have a common behaviour in url specification.

I spotted the [code responsible for feign client building](https://github.com/spring-cloud/spring-cloud-openfeign/blob/b13711f90655907b15874096348edd851b790c3e/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java#L357) and think that it would be easily doable by

- only appending ̀http://` if `url` attribute value does not match the `[a-z]+://.*` pattern
- remove load balancing client if and only if `url` attribute does not starts with `lb://`

I am eager to implement it if you allow me to do so. I am also pleased to discuss about it and about any alternative I missed.

Thank you

antechrestos:
Description
=========

Fixes #489 : This change is the one requested by late issue . The aim is to handle url such as `lb://some-service` (or `lbs://some-service). When such url is specified, we ensure  load balance client is provided and we keep it instead of using delegate one).

Code implementation
=================

I willingly kept 3 commit to better describe the implementation

- improve code specification of `name`, `url` and `path` handle by adding tests
- refactoring of code, particularly by forcing reentrance of code (`url`  was modified by side effect)
- implementing the feature


Sam-Kruglov:
Are you sure it's related to 499? Doesn't seem like it is to me...

antechrestos:
@Sam-Kruglov Yeah sorry. Thank you for the warning. :pray:

OlgaMaciaszek:
See https://github.com/spring-cloud/spring-cloud-openfeign/issues/489#issuecomment-817707750

Issue #542, opened 2021-05-12
Refactor Feign Hal feature to use Hateoas 1.3 augmentClient() method


OlgaMaciaszek:

OlgaMaciaszek:
Done.

PR #543, opened 2021-05-13
Use WebConverters for Hateoas.

OlgaMaciaszek:
Fixes gh-542.


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringDecoder.java
```diff
@@ -21,38 +21,56 @@
...
-	private ObjectFactory<HttpMessageConverters> messageConverters;
+	private final ObjectFactory<HttpMessageConverters> messageConverters;
 
+	private final ObjectProvider<HttpMessageConverterCustomizer> customizers;
+
+	/**
+	 * @deprecated in favour of
+	 * {@link SpringDecoder#SpringDecoder(ObjectFactory, ObjectProvider)}
+	 */
+	@Deprecated
```
nilson-aguiar, line 56: Hi @OlgaMaciaszek, just checking with you, does this constructor needs to be marked as `Deprecated`?

In my case, for example, I don't have any HttpMessageConverterCustomizer to create, so this constructor works fine for me, but I keep getting that a deprecated constructor is being used. 

I can't use the `EmptyObjectProvider` because it is package-private so I'd have to make my own to not use the default constructor and either can't use a null value on it.

Does it make sense to you to remove the `Deprecated` annotation from it?

PR #556, opened 2021-06-14
Simplify Log.info logic

michaldo:
'url' is never null in given case. At least I think so. Even if I'm wrong there is no need to distinguish between null and empty. Goal is simple: Inform reader that feign is created with load-balancer, not with url.


OlgaMaciaszek (approved):
LGTM.

Issue #564, opened 2021-06-30
Feign Inheritance has conflicts with the web mvc


terminux:
When I use [Feign Inheritance](https://docs.spring.io/spring-cloud-openfeign/docs/current/reference/html/#spring-cloud-feign-inheritance "Feign Inheritance"), I get the following errors:

```java
Caused by: java.lang.IllegalStateException: Ambiguous mapping. Cannot map 'xxxx.UserClient' method 
xxx.UserClient#getUser(Long)
to {GET [/v1/users/{id}]}: There is already 'userResource' bean method
xxx.UserResource#getUser(Long) mapped.
	at org.springframework.web.servlet.handler.AbstractHandlerMethodMapping$MappingRegistry.validateMethodMapping(AbstractHandlerMethodMapping.java:662)
```

Because I added the `@RequestMapping` annotation on the `UserService` class, just like that:

```java
@RequestMapping("/v1")
public interface UserService {

    @RequestMapping(method = RequestMethod.GET, value ="/users/{id}")
    User getUser(@PathVariable("id") long id);
}
```

So `Spring MVC` thinks that `UserClient` is also a handler,This will conflict with the `UserResource`.

Can I submit a pull request to fix this problem?

cbezmen:
@terminux 

Why don't you use path variable `@FeignClient(path = "/v1")` instead of @RequestMapping("/v1") ?

terminux:
@cbezmen 

> @terminux
> 
> Why don't you use path variable `@FeignClient(path = "/v1")` instead of `@RequestMapping("/v1")` ?

This is because I want to reuse the UserService interface

terminux:
@cbezmen Sorry, I closed the issue,by accident.

terminux:
> @terminux
> 
> Why don't you use path variable `@FeignClient(path = "/v1")` instead of `@RequestMapping("/v1")` ?

Because I used Feign Inheritance to reuse the `UserService` interface

terminux:
Sorry, I closed this issue by accident just now.

terminux:
> @terminux
> 
> Why don't you use path variable `@FeignClient(path = "/v1")` instead of `@RequestMapping("/v1")` ?

I used `Feign Inheritance` to reuse the `UserService` interface, because the code in `UserService` and `UserClient` is the same.

[...]

PR #565, opened 2021-06-30
Fix Feign Inheritance conflicts with Handler

terminux:
Fixes gh-564


OlgaMaciaszek:
There's an ongoing discussion in the issue.

OlgaMaciaszek:
Thanks for this PR, however will be closing it as this behaviour is by design (see discussion in gh-564).

Issue #576, opened 2021-07-20
Regex expression in path pattern is not supported


quaff:
**Describe the bug**
according to https://docs.spring.io/spring-framework/docs/3.2.x/spring-framework-reference/html/mvc.html#mvc-ann-requestmapping-uri-templates-regex, Spring MVC supports regex extraction for @PathVariable
```java
@GetMapping("/users/{id:\\d+}")
public User get(@PathVariable String id) {
}
```
but feign will throw IllegalStateException
```
Caused by: java.lang.IllegalStateException: Body parameters cannot be used with form parameters.
Warnings:
- 
	at feign.Util.checkState(Util.java:129)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:125)
	at org.springframework.cloud.openfeign.support.SpringMvcContract.parseAndValidateMetadata(SpringMvcContract.java:194)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:62)
	at feign.ReflectiveFeign$ParseHandlersByName.apply(ReflectiveFeign.java:151)
	at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:49)
	at feign.Feign$Builder.target(Feign.java:269)
	at org.springframework.cloud.openfeign.DefaultTargeter.target(DefaultTargeter.java:30)
	at org.springframework.cloud.openfeign.FeignClientFactoryBean.getTarget(FeignClientFactoryBean.java:419)
	at org.springframework.cloud.openfeign.FeignClientFactoryBean.getObject(FeignClientFactoryBean.java:371)
	at org.springframework.cloud.openfeign.FeignClientsRegistrar.lambda$registerFeignClient$0(FeignClientsRegistrar.java:235)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.obtainFromSupplier(AbstractAutowireCapableBeanFactory.java:1231)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1173)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:564)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:524)
	... 82 more
```

PR #577, opened 2021-07-20
Support regex expression for @PathVariable

quaff:
Fix gh-576

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
@@ -52,6 +52,9 @@ public interface StoreClient {
 
 	@RequestMapping(method = RequestMethod.POST, value = "/stores/{storeId}", consumes = "application/json")
 	Store update(@PathVariable("storeId") Long storeId, Store store);
+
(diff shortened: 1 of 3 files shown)


quaff:
Please review @OlgaMaciaszek

OlgaMaciaszek:
Fixes gh-576.

OlgaMaciaszek (changes requested):
@quaff Thanks for submitting the PR.  In general, the change looks good. However, please add information about the feature [to the docs](https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/docs/src/main/asciidoc/spring-cloud-openfeign.adoc).

quaff:
@OlgaMaciaszek It is bug fix, such feature in Spring MVC existed before Spring Cloud created, people should expect it works without checking the docs, WDYT?

OlgaMaciaszek:
@quaff  Feign Client is not the same as Spring MVC. We strive to provide integration for as many features as possible within SC OpenFeign, but it's a separate lib. It's never a good idea to add info to docs (just one sentence with a simple example will be fine).

OlgaMaciaszek (approved):
LGTM

PR #583, opened 2021-07-26
Respect SpringDataWebProperties where possible

quaff:
`PageableSpringQueryMapEncoder` should use parameter names from `SpringDataWebProperties` if customized as well as `PageableSpringEncoder`

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
@@ -71,6 +71,7 @@ import static feign.form.ContentType.MULTIPART;
  * @author Jonatan Ivanov
  * @author Olga Maciaszek-Sharma
  * @author Hyeonmin Park
+ * @author Yanming Zhou
  */
 @Configuration(proxyBeanMethods = false)
 public class FeignClientsConfiguration {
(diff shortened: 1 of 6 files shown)


spencergibb (changes requested):
Please add a test

quaff:
> Please add a test

@spencergibb test added, and `PageableEncoderTests` revised to cover `SpringDataWebProperties` supports.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/PageableEncoderTests.java
```diff
@@ -37,10 +37,13 @@
  * Tests the pagination encoding and sorting.
  *
  * @author Charlie Mordant.
+ * @author Yanming Zhou
  */
```
OlgaMaciaszek, line 41: Please also add a test that verifies the default param names are used if not redefined in properties.

OlgaMaciaszek (changes requested):
@quaff Thanks for submitting the PR. Looks good. There are just some minor items to change:
- have added a comment - please address
- please add `@author` tag with your full name to javadocs of all the classes you've modified

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/PageableEncoderTests.java
quaff: Done @OlgaMaciaszek

OlgaMaciaszek (approved):
LGTM

Issue #563, opened 2021-06-28
Switch remaining jUnit 4 tests to jUnit jupiter


OlgaMaciaszek:

Syemon:
Hi, I would like to work on this issue.

OlgaMaciaszek:
Hi @Syemon - great :) . Just please let us know if you will be converting the entire project or just some specific modules or packages.

OlgaMaciaszek:
@Syemon Please add your changes against the `3.0.x` branch.

OlgaMaciaszek:
@Syemon Please let me know what your planned timeline is for submitting this PR.

Syemon:
@OlgaMaciaszek I plan submitting this PR on Monday 02.08

Syemon:
@OlgaMaciaszek I have created draft because I am not sure about one test.

PR #588, opened 2021-08-01
Switch remaining jUnit4 tests to jUnit jupiter

Syemon:
Fixes [#563]
I am not sure about this test. It passes if I run it directly through maven command, but It fails with ```org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.apache.http.impl.client.CloseableHttpClient' available``` if I run it through inteliJ.
```java
@Test
public void hc4ShouldWinEvenHc5ConfigEnabled() {
	ConfigurableApplicationContext context = new SpringApplicationBuilder()
			.properties("feign.httpclient.hc5.enabled=true").web(WebApplicationType.NONE)
			.sources(HttpClientConfiguration.class, FeignAutoConfiguration.class).run();

	verifyHc4BeansAvailable(context);

	if (context != null) {
		context.close();
	}
}
```


OlgaMaciaszek:
@Syemon we treat mvn runs as the source of truth. Might be some issue with IntelliJ. Does this test behave differently with your changes than without them when you run it from IDE (i.e. was it passing in the IDE before your changes or not?)?

Syemon:
@OlgaMaciaszek it Was passing before change.

OlgaMaciaszek:
Ok. Will check that.

OlgaMaciaszek:
@Syemon that test works fine for me, both with Maven and through IDE, so it's probably just a local issue. We can be disregarded. Please resubmit your changes against the `main` branch (merge in any newly added code from `main` to your branch first) - will review the whole PR then.

Syemon:
Thanks, I was not aware. I will make a new branch just for my changes.

Issue #607, opened 2021-09-28
Cache support


Sam-Kruglov:
I wish there was out-of-the-box support for `@Cacheable` on feign clients.

Here's how I made it work:
```java
import feign.Capability;
import feign.InvocationHandlerFactory;
import feign.InvocationHandlerFactory.MethodHandler;
import feign.Target;
import org.aopalliance.intercept.MethodInvocation;
import org.springframework.cache.interceptor.CacheInterceptor;
import org.springframework.stereotype.Component;

import java.lang.reflect.AccessibleObject;
import java.lang.reflect.Method;
import java.util.Map;

import static java.util.stream.Collectors.toMap;

/**
 * Allows Spring's @Cache* annotations to be declared on the feign client's methods
 */
@Component
public class CachingCapability implements Capability {
    private final CacheInterceptor cacheInterceptor;

    public CachingCapability(CacheInterceptor cacheInterceptor) {
        this.cacheInterceptor = cacheInterceptor;
    }

    @Override
    public InvocationHandlerFactory enrich(InvocationHandlerFactory invocationHandlerFactory) {
        return (target, dispatch) -> invocationHandlerFactory.create(target, withCacheProxy(target, dispatch));
    }

    private Map<Method, MethodHandler> withCacheProxy(Target<?> target, Map<Method, MethodHandler> dispatch) {
        return dispatch.entrySet().stream()
                .map(e -> Map.entry(e.getKey(), (MethodHandler) argv -> cacheInterceptor.invoke(new MethodInvocation() {
                    @Override
                    public Method getMethod() {
                        return e.getKey();
                    }

                    @Override
                    public Object[] getArguments() {
                        return argv;
                    }

                    @Override
                    public Object proceed() throws Throwable {
                        return e.getValue().invoke(argv);
                    }

                    @Override
                    public Object getThis() {
                        return target;
                    }

                    @Override
                    public AccessibleObject getStaticPart() {
                        return e.getKey();
                    }
                }))).collect(toMap(Map.Entry::getKey, Map.Entry::getValue));
    }
}
```

Also see: https://github.com/OpenFeign/feign/issues/1513

PR #608, opened 2021-09-28
Cache support

Sam-Kruglov:
This is my approximate incomplete solution. Let me know your thoughts on the approach.

see #607
see https://github.com/OpenFeign/feign/pull/1515 (uses unreleased code from here)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -105,6 +108,12 @@ public FeignContext feignContext() {
 		return context;
 	}
 
+	@Bean
+	@ConditionalOnBean(CacheInterceptor.class)
+	public MethodHandlerCustomizer cacheHandlerCustomizer(CacheInterceptor cacheInterceptor) {
+		return MethodInterceptorAdapter.asCustomizer(cacheInterceptor);
+	}
```
Sam-Kruglov, line 115: This only covers `CacheInterceptor` but any `MethodInterceptor` would work here. I'm not strong in AOP, so maybe someone has an idea how to plug all of the interceptors here naturally.

The annotations on the feign client are on the interface. `MethodHandlerCustomizer` provides a `feign.Target` object, which is backing object that actually serves the request. `target.getType()` returns the interface itself. `MethodHandlerCustomizer` also provides a `java.lang.reflect.Method` which belongs to the interface too.

OlgaMaciaszek:
@Sam-Kruglov Thanks for the PR. Will discuss it with the team and get back to you next week. For starters, please revert the pom change as it should not be there.

OlgaMaciaszek:
Also, @Sam-Kruglov what was the reason for going with the `MethodHandlerCustomizer` in this PR in place of using the `Capability` as proposed in [this draft](https://github.com/spring-cloud/spring-cloud-openfeign/issues/607)?

Sam-Kruglov:
@OlgaMaciaszek I got feedback, forget the other PR, should be doing it via invocation handlers like [MeteredInvocationHandleFactory](https://github.com/OpenFeign/feign/blob/master/micrometer/src/main/java/feign/micrometer/MeteredInvocationHandleFactory.java). This repo also has a similar class but on a lower level [FeignCircuitBreakerInvocationHandler](https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java)

OlgaMaciaszek:
Fixes gh-607.

selvait:
Hi,
On seeing multiple related issues for cache support, I am not seeing any further updates on official release of this functionality. 
Can someone clarify if this cache support work is still in progress or released already? If release, kindly share me the feign version to use. Thanks.

selvait:
Thanks for the response @OlgaMaciaszek . By saying that PR were merged, I assume that cache support functionality is officially released. 
I was checking release notes of all the version from feignClient 11.0 to the latest one, but I dont see any mention on this cache support release. If released, can you kindly share the spring cloud version carrying this change. 

Thanks in advance.

Issue #594, opened 2021-08-16
Programmatic custom context inside FeignContext


Sam-Kruglov:
I can do `@FeignClient(contextId = "fooClient", name = "stores", configuration = FooConfiguration.class)`.
But if I create the clients manually via FeignClientBuilder (uses ApplicationContext) then I can't see a way to override stuff from FeignClientsConfiguration like an encoder.

I see that there is a FeignContext class, and the concept of NamedContextFactory but there is no way to use it, as I don't have access to the getContext method to register my stuff, nor to implement my own FeignClientSpecification.
Since it's there already, I don't think it's a feature request.

My concrete use case is that I have several vendor integrations at the same time, and each vendor has several feign clients with different credentials. One vendor has a custom encoder, and each has a custom decoder to throw custom exceptions. I create each client at runtime by using a FeignClientBuilder. It would've been a little cleaner if I extracted these customizations into FeignClientSpecification for each vendor rather than having them in code.

OlgaMaciaszek:
HEllo, @Sam-Kruglov , could you please specify with a code example what exactly you would like to do in your use-case and what methods would be helpful so that we can consider whether we should make them available?

Sam-Kruglov:
Basically I want to provide my own FeignClientSpecification implementation but currently it’s not possible.
I’ll provide an example later, on a trip

spring-cloud-issues:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

Sam-Kruglov:
First, I define this:

```java
public class CustomFeignConfig {

    @Bean
    public XmlMapper xmlMapper() {
        return XmlMapper.builder()
                .findAndAddModules()
                .propertyNamingStrategy(UPPER_CAMEL_CASE)
                .visibility(FIELD, ANY)
                .visibility(GETTER, NONE)
                .visibility(IS_GETTER, NONE)
                .build();
    }

    @Bean
    public ErrorDecoder errorDecoder(XmlMapper mapper) {
        return new CustomErrorDecoder(mapper);
    }

    @Bean
    public BeanPostProcessor delegateEncoder(XmlMapper mapper){
        return new BeanPostProcessor() {
            @Override
            public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
                if (bean instanceof Encoder) {
                    return new XmlDelegateEncoder(mapper, (Encoder) bean);
                }
                return bean;
[...]

PR #609, opened 2021-09-29
make FeignClientSpecification public

Sam-Kruglov:
Cannot be used otherwise


spencergibb:
Used for what?

Sam-Kruglov:
@spencergibb for this. It allows to declare classes with `@Bean` that will only affected a named context. See #594 
https://github.com/spring-cloud/spring-cloud-openfeign/blob/1ff68c11d241bc2ba7f52553a1b1b60e64b7a8c7/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java#L93-L94

spencergibb:
I'm sorry, I don't understand that. Care to share a sample of what you mean?

Sam-Kruglov:
@spencergibb this functionality already exists, here are some tests I found https://github.com/spring-cloud/spring-cloud-openfeign/blob/1ff68c11d241bc2ba7f52553a1b1b60e64b7a8c7/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientFactoryTests.java#L57

It basically allows to override stuff that is declared in [FeignClientsConfiguration](https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java). [FeignContext](https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignContext.java) uses it as "default" and then takes other beans from the spring context, but the context is separated and has a name.

Sam-Kruglov:
See #594

spencergibb:
That's already possible without making this public

Sam-Kruglov:
If you mean using feign's `Capability`, then that's different. Is it really possible to declare a class with `@Bean` methods that will only affect a specific `contextId` of `FeignContext`? There are no other ways to create `FeignClientSpecification` other than using its constructor.

Sam-Kruglov:
What can this field possibly autowire if these can't be registerred? 
https://github.com/spring-cloud/spring-cloud-openfeign/blob/1ff68c11d241bc2ba7f52553a1b1b60e64b7a8c7/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java#L93-L94

spencergibb:
It's been an internal implementation for many years

Sam-Kruglov:
One more note: for annotation-based approach there is `@FeignClient#configuration` but there isn't anything for programmatic approach

spencergibb:
This is what I was looking for

> for annotation-based approach there is @FeignClient#configuration but there isn't anything for programmatic approach

OlgaMaciaszek:
Fixes gh-594.

Issue #629, opened 2021-11-20
Java 16+ Records as `@SpringQueryMap` does not recognize any field


KENNYSOFT:
**Is your feature request related to a problem? Please describe.**
If you use Records in Java 16+ ([JEP 395](https://openjdk.java.net/jeps/395)) as `@SpringQueryMap`, it does not recognize any field.

**Describe the solution you'd like**

Create src/main/java/org/springframework/cloud/openfeign/support/RecordSpringQueryMapEncoder.java:

```java
// Skipped License and imports part.

/**
 * Provides support for encoding Record annotated as
 * {@link org.springframework.cloud.openfeign.SpringQueryMap}.
 *
 * @author Hyeonmin Park
 * @since Maybe 3.0.6
 */
public class RecordSpringQueryMapEncoder extends BeanQueryMapEncoder {

	@Override
	public Map<String, Object> encode(Object object) {
		if (supports(object)) {
			try {
				RecordComponent[] component = object.getClass().getRecordComponents();
				Map<String, Object> propertyNameToValue = new HashMap<>();
				for (RecordComponent rc : component) {
					Method method = rc.getAccessor();
					Object value = method.invoke(object);
					if (value != null && value != object) {
						String name = rc.getName();
						propertyNameToValue.put(name, value);
					}
				}
				return propertyNameToValue;
			}
			catch (IllegalAccessException | InvocationTargetException e) {
				throw new EncodeException("Failure encoding object into query map", e);
			}
		}
		else {
			return super.encode(object);
		}
	}

	protected boolean supports(Object object) {
		return object.getClass().isRecord();
	}

}
```

Add a bean to src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java:

```java
@Bean
@ConditionalOnJava(JavaVersion.SIXTEEN)
@ConditionalOnMissingBean
public QueryMapEncoder feignQueryMapEncoderRecord() {
	return new RecordSpringQueryMapEncoder();
}
```

But in this way, we need to only build with Java 16+, and I think it can break many compatibilities. Is there any way to include some classes using the Records feature with Java 8? or just by splitting the artifact something like 'spring-cloud-openfeign-jdk16'?

**Describe alternatives you've considered**
* I can implement this in my project, but it's better to provide a common solution in the library.
* I'm also considering making support for Records directly in https://github.com/OpenFeign/feign but still unfamiliar with.

**Additional context**
[...]

PR #630, opened 2021-11-24
Add queryMapEncoder to FeignClientProperties

KENNYSOFT:
From https://github.com/spring-cloud/spring-cloud-openfeign/issues/629#issuecomment-976496995

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-openfeign.adoc
@@ -192,6 +192,7 @@ feign:
 				capabilities:
 					- com.example.FooCapability
 					- com.example.BarCapability
+				queryMapEncoder: com.example.SimpleQueryMapEncoder
 				metrics.enabled: false
 ----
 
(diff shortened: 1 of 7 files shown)


OlgaMaciaszek (approved):
@KENNYSOFT Thanks for submitting the PR. It looks fine, however, more docs should have been added. Since we're planning to release `2021.0.1` today, I'm going to add them, in order to get this into the release.

OlgaMaciaszek:
After a further consideration, although the code looks good, I don't see that it's a needed functionality - the general model of overriding these Feign configuration elements is via adding Java beans as also described in the documentation. Is there a scenario where that is causing an issue? 
I'm going to leave this PR open to get opinions from the community, however, if no compelling reasons will be presented to add this, we'll stay just with the bean config option.

KENNYSOFT:
Actually, `Retryer`, `ErrorDecoder`, and `Capability` are also supported via Java Bean with the way in application.yml. There are not any 'compelling reasons' but it can give more flexibility to set the `QueryMapEncoder`.

KENNYSOFT:
Just found out fe8bbbb6dab491fb9d14e2946cf41af6b096e73a and integrated since v3.1.1; don't care whether i it included in v3.0.x or not for now. Closing with thanks!

PR #636, opened 2021-11-28
Support size = 0 in PageJacksonModule with test

nkonev:
From https://github.com/spring-cloud/spring-cloud-openfeign/pull/592

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/PageJacksonModule.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/PageJacksonModule.java
@@ -38,6 +38,8 @@ import org.springframework.data.domain.Sort;
  *
  * @author Pascal Büttiker
  * @author Olga Maciaszek-Sharma
+ * @author Pedro Mendes
+ * @author Nikita Konev
  */
 public class PageJacksonModule extends Module {
 
@@ -69,18 +71,22 @@ public class PageJacksonModule extends Module {
 				@JsonProperty("size") int size, @JsonProperty("totalElements") @JsonAlias({ "total-elements",
 						"total_elements", "totalelements", "TotalElements" }) long totalElements,
 				@JsonProperty("sort") Sort sort) {
-			PageRequest pageRequest;
-			if (sort != null) {
-				pageRequest = PageRequest.of(number, size, sort);
+			if (size > 0) {
+				PageRequest pageRequest;
+				if (sort != null) {
+					pageRequest = PageRequest.of(number, size, sort);
+				}
+				else {
+					pageRequest = PageRequest.of(number, size);
+				}
+				delegate = new PageImpl<>(content, pageRequest, totalElements);
 			}
 			else {
-				pageRequest = PageRequest.of(number, size);
+				delegate = new PageImpl<>(content);
 			}
-			delegate = new PageImpl<>(content, pageRequest, totalElements);
-
 		}
 
-		@JsonProperty
+		@JsonIgnore
 		@Override
 		public int getTotalPages() {
 			return delegate.getTotalPages();
@@ -104,7 +110,7 @@ public class PageJacksonModule extends Module {
 			return delegate.getSize();
 		}
 
-		@JsonProperty
+		@JsonIgnore
 		@Override
 		public int getNumberOfElements() {
 			return delegate.getNumberOfElements();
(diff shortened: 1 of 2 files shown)


OlgaMaciaszek (approved):
@nkonev Thanks for fixing this up. Looks good. However, I'm going to cherry-pick this commit directly to `3.0.x`, so that this is also included in the `2020.0.x` release train.

OlgaMaciaszek:
Merged with ad892a23f0d317dd9afe0cf63460122b3ff02780.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/PageJacksonModule.java
```diff
@@ -69,18 +71,22 @@ public void setupModule(SetupContext context) {
...
+				delegate = new PageImpl<>(content);
 			}
-			delegate = new PageImpl<>(content, pageRequest, totalElements);
-
 		}
 
-		@JsonProperty
+		@JsonIgnore
 		@Override
 		public int getTotalPages() {
```
KENNYSOFT, line 91: Was there any reason to add `@JsonIgnore` annotation to such properties? I'm facing the property missing problem, since we just get `Page<T>` instance from the FeignClient and then return it to the response.

nkonev: I ignored them because ignored properties didn't deserialized correctly, try un-ignore them and run PageJacksonModuleTests.

KENNYSOFT: Thanks for the fast reply. I've confirmed that it says:

> com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "last" (class org.springframework.cloud.openfeign.support.PageJacksonModule$SimplePageImpl), not marked as ignorable (5 known properties: "content", "totalElements", "sort", "number", "size"])
 at [Source: (String)"{"content":[],"pageable":"INSTANCE","totalElements":0,"last":true,"totalPages":1,"size":0,"number":0,"sort":[],"numberOfElements":0,"first":true,"empty":true}"; line: 1, column: 132] (through reference chain: org.springframework.cloud.openfeign.support.PageJacksonModule$SimplePageImpl["last"])

But I don't think changing `@JsonProperty` to `@JsonIgnore` is a good choice. Instead, we'd better disable `FAIL_ON_UNKNOWN_PROPERTIES` from the `objectMapper`, since the `PageJacksonModule` should be able to deserialize from any other form of `Page<T>` JSON.

We can see the test passes with un-ignore and then disable the feature as mentioned:

```java
objectMapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```

@OlgaMaciaszek Can I get your opinion?

nkonev: @KENNYSOFT Personally I would prefer to use `@JsonIgnoreProperties(ignoreUnknown = true)`
Like here https://github.com/spring-projects/spring-security/blob/ad907457eeeee46da101215f5408d3cd98de4881/web/src/main/java/org/springframework/security/web/jackson2/DefaultCsrfTokenMixin.java#L39

Modification of "global" objectMapper is not good choice because many other parts use it.

Can you provide reproducer? In order to include it to test.

I tried "cascase" serialization-deserialization-serializaton and it failed.

KENNYSOFT: Actually, I've modified a local objectMapper, especially https://github.com/spring-cloud/spring-cloud-openfeign/pull/636/files#diff-6338018e154894301cfd5d80b35fd208bf2ce3a6387c3d829d5e2cbc60edcbccR48 here.

nkonev: https://github.com/spring-cloud/spring-cloud-openfeign/pull/653

OlgaMaciaszek:
Hi, @KENNYSOFT Thanks for raising it. I agree with @nkonev that `@JsonIgnoreProperties(ignoreUnknown = true)` is a better solution.

Issue #671, opened 2022-01-25
It's not possible to provide a custom Feign.Client when instantiating programatically


f11h:
**Describe the bug**
When using the ```.customize()``` method of ```FeignClientBuilder``` it is not possible to provide a custom (ApacheHttp)Client for request executing. The set client will be overriden *after* the customizer was apllied.

**Sample**

We have one Interface with definition of the REST Endpoints. We want to use this Interface to contact multiple instances of this service at different URLs. For all of these connections we need seperate HTTPS Client Certificates and Truststores. This configuration needs to be done at runtime when instantiating the Feign Client.

Example Code of Client instantiation:

```java

Feign.Client client = myCustomClientWithCustomSslContext(clientCert, privateKey, trustStore);
String url = "https://example.org";

MyRestClient myRestClient = new FeignClientBuilder(context)
            .forType(MyRestClient.class, "new-context-id")
            .customize(builder -> builder.client(client))
            .url(url)
            .build();
```

The FeignClient will be created, but the self created (ApacheHttp)Client with the mutual TLS configuration will be replaced by a default client created when the FeignContext gets initialized.

This happens at ```FeignClientFactoryBean#getTarget()``` (https://github.com/spring-cloud/spring-cloud-openfeign/blob/261fc58a5ef0d6f651e66ebca4d958167ebec942/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java#L409). We set our custom ApacheHttpClient as a customizer. The customizers are applied within this method call: https://github.com/spring-cloud/spring-cloud-openfeign/blob/261fc58a5ef0d6f651e66ebca4d958167ebec942/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java#L412

At this moment the ApacheHttpClient is set to the builder as expected. But a few lines later it will be overriden with the default one:
https://github.com/spring-cloud/spring-cloud-openfeign/blob/261fc58a5ef0d6f651e66ebca4d958167ebec942/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java#L432-L444

Maybe there is a workaround to archive what a want at another way, but at the moment I don't have any clue. I will provide a PR to adress this issue in a few hours.

OlgaMaciaszek:
@f11h Thanks for reporting this issue. I was able to reproduce it.

PR #672, opened 2022-01-25
Allow to inject custom HttpClient to programmatically created FeignClients

f11h:
The call of applyBuildCustomizers has been done before applying the HttpClient from FeignContext.

This PR moves the call at the end of the getTarget Method. It will allow to use a custom HttpClient. See #671 for more details.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
@@ -132,7 +132,6 @@ public class FeignClientFactoryBean
 		// @formatter:on
 
 		configureFeign(context, builder);
-		applyBuildCustomizers(context, builder);
 
 		return builder;
 	}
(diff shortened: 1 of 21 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
```diff
@@ -444,6 +443,7 @@ public Object getObject() {
 			builder.client(client);
 		}
 		Targeter targeter = get(context, Targeter.class);
+		applyBuildCustomizers(context, builder);
```
OlgaMaciaszek, line 446: Let's move this one line up.

OlgaMaciaszek (changes requested):
Hello, @f11h Thanks for submitting the PR. Have added a comment, please address. Also:
- please submit the fix against the `3.1.x` instead of `main`, so that the fix is also available for the `2021` release train
- please provide a test for the fix
- please add your full name and surname with `@author` tag to all the classes you've modified.

OlgaMaciaszek:
Fixes gh-671.

f11h:
I will create a new PR based on 3.1.x

PR #674, opened 2022-01-27
Allow to inject custom HttpClient to programmatically created FeignClients

f11h:
The call of applyBuildCustomizers has been done before applying the HttpClient from FeignContext.
This PR moves the call at the end of the getTarget Method. It will allow to use a custom HttpClient. See #671 for more details.

Follow Up PR for #672 (rebased on 3.1.x branch)

Move Call of ```applyBuildCustomizers()``` at end of ```getTarget()``` method.
Add Unit Test to verify custom HttpClients are set.
Modify Unit Tests to use ```getTarget()``` method instead of ```feign()``` method. ```feign()``` method has protected access and is only used within test scope. All other access happens through ```getTarget()``` method

fixes gh-671

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
@@ -69,6 +69,7 @@ import org.springframework.util.StringUtils;
  * @author Sam Kruglov
  * @author Jasbir Singh
  * @author Hyeonmin Park
+ * @author Felix Dittrich
  */
 public class FeignClientFactoryBean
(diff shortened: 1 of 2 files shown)


OlgaMaciaszek:
Fixes gh-671.

OlgaMaciaszek (approved):
Thanks, @f11h. LGTM.

christophejan:
@OlgaMaciaszek @f11h : Hello ! Just reading the fix, I think maybe something was missed : `getTarget` method has two `return` ; One for specified url, the other for discovery/loadbalancing. When I read the fix, I think that customizers will not be applied anymore when url is not specified. Is it an expected change ?

OlgaMaciaszek:
Hello, @christophejan good find :). Thanks for pointing this out. Will fix it.

OlgaMaciaszek:
Fixed with e99dac2af1217f02df2c56c490c7f4882c028464.

PR #687, opened 2022-03-09
Provide an alternate naming convention for CB ids

ryanjbaxter:
Configuration properties cannot contain characters like hash, parens, or commas.

See https://github.com/spring-cloud/spring-cloud-circuitbreaker/issues/141


spencergibb (approved):
LGTM, only thing is the name. Since circuit breaker is already part of the property, maybe `feign.circuitbreaker.alphanumericNameResolver.enabled` or `feign.circuitbreaker.alphanumericNames.enabled`

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -166,10 +166,20 @@ public Targeter defaultFeignTargeter() {
 
 		@Bean
 		@ConditionalOnMissingBean(CircuitBreakerNameResolver.class)
+		@ConditionalOnProperty(value = "feign.circuitbreaker.alphanumericNames.enabled",
```
spencergibb, line 169: We need to use - rather than camel case, so `alphanumeric-names.`

spencergibb (approved):

OlgaMaciaszek (approved):
LGTM.

PR #707, opened 2022-04-28
Delete '/' at the end of url

artem-gorshkov:
In situation when we create FeignClient in this way:
@FeignClient(url = "https://localhost/", path = "api/v2")
We will get an error because of adding '/' in the beginning of the path in FeignClientsRegistrar.getPath, result will:
https://localhost//api/v2
To escape this situation we need to delete '/' at the end of url in FeignClientsRegistrar.getUrl


gorshik:
@pivotal-cla This is an Obvious Fix

spencergibb:
Bug fixes like this don't typically qualify as an obvious fix

OlgaMaciaszek (changes requested):
Hello @artem-gorshkov, thank you for submitting the PR. Please add a test that verifies your fix.

OlgaMaciaszek:
Closing in favour of https://github.com/spring-cloud/spring-cloud-openfeign/pull/1100.

PR #713, opened 2022-05-10
More friendly hints in dynamic URL.

### LINKED ISSUE #712: More friendly hints in dynamic URL

> spring-cloud-openfeign-core:2.2.5.RELEASE

When I use a dynamic URL, like this:
```java
	@FeignClient(name = "dynamic")
	interface DynamicClient {

		@RequestLine("GET /")
		String get(URI uri);
	}
```
I will get an exception message:
```java
Caused by: com.netflix.client.ClientException: Load balancer does not have available server for client: github.com
	at com.netflix.loadbalancer.LoadBalancerContext.getServerFromLoadBalancer(LoadBalancerContext.java:483)
```

More friendly tips.

Internally provide a constant for the empty URL:

```java
	@FeignClient(name = "dynamic", url = FeignUtils.EMPTY_CLIENT, configuration = FooConfiguration.class)
	interface DynamicClient {

		@RequestLine("GET /")
		String get(URI uri);
	}
```
Also, add a comment for `url()`

```java
	/**
	 * @return an absolute URL or resolvable hostname (the protocol is optional).
	 * If you are using {@link java.net.URI}, configure the value to {@link FeignUtils#EMPTY_CLIENT}
	 */
	String url() default "";
```

crossoverJie:
fix #712

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClient.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClient.java
@@ -23,6 +23,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.springframework.cloud.openfeign.support.FeignUtils;
 import org.springframework.core.annotation.AliasFor;
 
 /**
@@ -98,6 +99,7 @@ public @interface FeignClient {
 
 	/**
 	 * @return an absolute URL or resolvable hostname (the protocol is optional).
+	 * If you are using {@link java.net.URI}, configure the value to {@link FeignUtils#EMPTY_CLIENT}.
 	 */
 	String url() default "";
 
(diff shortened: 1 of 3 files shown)


crossoverJie:
When I used it this way it never worked as expected, and after looking up a lot of information and source code I found that it only needed the url not to be equal to the empty string to work.

But the official comments don't explain much, so this will help other developers who are careless like me.

```java
	@FeignClient(name = "dynamic")
	interface DynamicClient {

		@RequestLine("GET /")
		String get(URI uri);
	}
```

crossoverJie:
Of course, please run this unit test directly to reproduce it.


https://github.com/crossoverJie/spring-cloud-openfeign/blob/c3c6973451dbbb3b7b43d9ad6b41d8e498c41cc7/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignClientErrorDecoderTests.java#L74

OlgaMaciaszek:
Hello,  @crossoverJie; thanks for submitting a PR. The `2.2.x` branch is no longer supported. The test you've provided does not reproduce a valid use-case. Also, it lacks any assertion. Apart from this, I don't think we should be adding a `String` constant just for the benefit of linking to it from the javadoc.

Issue #749, opened 2022-09-08
connecttimeout and readtimeout will be reset when client has any configuration


izdt:
**Describe the bug**
After the bug ([Read timeout from application properties is not used when client is configured ](https://github.com/spring-cloud/spring-cloud-openfeign/issues/324) ) was fixed in [v2.2.4.RELEASE](https://github.com/spring-cloud/spring-cloud-openfeign/releases/tag/v2.2.4.RELEASE), connect timeout and readtimeout of [LoadBalancerFeignClient](https://github.com/spring-cloud/spring-cloud-openfeign/blob/2.2.x/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/ribbon/LoadBalancerFeignClient.java#L97) cannot work when client has any configuration.

**Sample**
When feign client not config, the readtimeout and connecttimeout load by IClientConfig is 1000ms / 1000ms (default, or configed to others ) ( see LoadBalancerFeignClient#L97)
```
//LoadBalancerFeignClient#L97
IClientConfig getClientConfig(Request.Options options, String clientName) {
		IClientConfig requestConfig;
		if (options == DEFAULT_OPTIONS) {
			requestConfig = this.clientFactory.getClientConfig(clientName);
		}
		else {
			requestConfig = new FeignOptionsClientConfig(options);
		}
		return requestConfig;
	}
```
and 
```
//FeignClientFactoryBean with bug 323:
if (config.getConnectTimeout() != null && config.getReadTimeout() != null) {
			builder.options(new Request.Options(config.getConnectTimeout(),
					config.getReadTimeout()));
		}
``` 
It was worked before bug 323 fixed in [FeignClientFactoryBean.java](https://github.com/spring-cloud/spring-cloud-openfeign/commit/72784920d8bb1b2bae32fee923aed154e2372531#diff-a0977560e79ed9a80e77ff642408f605815d8eed4100a5f6aff71087743b317f). But now when client has any configuration,  readtimeout and connecttimeout (1000ms/1000ms) will be reset to 10000ms and 60000ms. For example, put a Logger leave in properties:
```
feign.client.config.default.loggerLevel=FULL
```

OlgaMaciaszek:
Hello, @izdt  `2.2.x` is no longer supported. Could you please verify if the issue can be reproduced on a supported version (`3.1.4`, Spring Cloud dependencies `2021.0.4`) and let me know here?

[...]

PR #750, opened 2022-09-08
Add back Check to avoid connectTimeout and readTimeout  being reset

izdt:
Fix issue [749](https://github.com/spring-cloud/spring-cloud-openfeign/issues/749)


OlgaMaciaszek:
Closing since not active since 2022.

PR #751, opened 2022-09-09
Disable loadbalancer clients retry when clients retry is disabled

WonSik36:
Hi, I wrote this PR because the load balancer client retry properties did not work as expected.

I have loadbalancer properties as below (1 default, 1 individual clients)

```yml
spring.cloud.loadbalancer:
  retry:
    enabled: true
    retryable-status-codes: 503
    max-retries-on-next-service-instance: 1

  clients:
    retry-disabled-service:
      retry:
        enabled: false
```

And I expect when I execute `retry-disabled-service` client it should not retry.
However, the client will do the retry as set in default properties.

I think the `enabled` property of that individual client retry should be checked. [reference](https://docs.spring.io/spring-cloud-commons/docs/current/reference/html/#retrying-failed-requests)
And I checked `spring.cloud.loadbalancer.clients.<clientId>.retry.enabled` is not checked when set `RetryPolicy` to `RetryTemplate`.

Therefore, I made this pull request with reference to [RetryLoadBalancerInterceptor#createRetryTemplate](https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/RetryLoadBalancerInterceptor.java#L163).
Plus, I added test, too.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/RetryableFeignBlockingLoadBalancerClient.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/loadbalancer/RetryableFeignBlockingLoadBalancerClient.java
@@ -231,7 +231,9 @@ public class RetryableFeignBlockingLoadBalancerClient implements Client {
 			retryTemplate.setListeners(retryListeners);
 		}
 
-		retryTemplate.setRetryPolicy(retryPolicy == null ? new NeverRetryPolicy()
+		retryTemplate.setRetryPolicy(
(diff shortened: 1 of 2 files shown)


OlgaMaciaszek (changes requested):
Hello @WonSik36 , thanks a lot for submitting this PR. Sorry for not getting back to you earlier. This makes sense. Please add your full name with `@author` tag to the classes you've modified. Also, could you please submit these changes against `3.1.x` instead of `main` so that we can include them starting with the `2021.x` release line.

WonSik36:
@OlgaMaciaszek 
Thanks for your kind reply.
I wrote a new PR by your requests. #773 
And I will close this PR.

PR #754, opened 2022-09-19
Sets up MicrometerCapabilities

marcingrzejszczak:
prerequisite - https://github.com/OpenFeign/feign/pull/1760
sample - https://github.com/micrometer-metrics/micrometer-samples/commit/7485ae550d197d6c205f660a57545b037b59c554

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
@@ -28,7 +28,9 @@ import feign.codec.Decoder;
 import feign.codec.Encoder;
 import feign.form.MultipartFormContentProcessor;
 import feign.form.spring.SpringFormEncoder;
+import feign.micrometer.FeignObservationConvention;
 import feign.micrometer.MicrometerCapability;
+import feign.micrometer.ObservedClientInterceptor;
 import feign.optionals.OptionalDecoder;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.observation.ObservationRegistry;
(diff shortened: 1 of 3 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/annotation/QueryMapParameterProcessor.java
```diff
@@ -45,7 +45,7 @@ public boolean processArgument(AnnotatedParameterContext context, Annotation ann
 		MethodMetadata metadata = context.getMethodMetadata();
 		if (metadata.queryMapIndex() == null) {
 			metadata.queryMapIndex(paramIndex);
-			metadata.queryMapEncoded(SpringQueryMap.class.cast(annotation).encoded());
+			// metadata.queryMapEncoded(SpringQueryMap.class.cast(annotation).encoded());
```
marcingrzejszczak, line 48: This is no longer there in 11.10-SNAPSHOT

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -243,8 +244,10 @@ protected static class MetricsConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
-		public MicrometerCapability micrometerCapability(MeterRegistry meterRegistry) {
-			return new MicrometerCapability(meterRegistry);
+		public MicrometerCapability micrometerCapability(MeterRegistry meterRegistry,
```
jonatan-ivanov, line 247: I'm thinking if we should add flags so that users can switch back to the `MicrometerCapability` if they want the old behavior.

jonatan-ivanov (approved):

OlgaMaciaszek:
Closing as the corresponding feature has already been added.

Issue #644, opened 2021-12-09
Migrate from Spring Security OAuth to OAuth2 support in Spring Security


OlgaMaciaszek:

xenoterracide:
is this what's causing this?

```
   FeignAutoConfiguration.Oauth2FeignConfiguration:
      Did not match:
         - @ConditionalOnClass did not find required class 'org.springframework.security.oauth2.client.OAuth2ClientContext' (OnClassCondition)
 ```

```
❯ fd gradle.lockfile -x grep oauth {} | awk -F'=' '{ print $1 }' | sort -u
com.nimbusds:oauth2-oidc-sdk:9.35
org.springframework.boot:spring-boot-starter-oauth2-client:2.7.3
org.springframework.security:spring-security-oauth2-client:5.7.3
org.springframework.security:spring-security-oauth2-core:5.7.3
org.springframework.security:spring-security-oauth2-jose:5.7.3
❯ fd gradle.lockfile -x grep feign {} | awk -F'=' '{ print $1 }' | sort -u
io.github.openfeign.form:feign-form-spring:3.8.0
io.github.openfeign.form:feign-form:3.8.0
io.github.openfeign:feign-core:11.8
io.github.openfeign:feign-slf4j:11.8
org.springframework.cloud:spring-cloud-openfeign-core:3.1.4
org.springframework.cloud:spring-cloud-starter-openfeign:3.1.4
```

xenoterracide:
if not, should I report another bug here? or where? I can't seem to get oauth2 configured per docs

OlgaMaciaszek:
@xenoterracide The migration has been done - will be available with the latest release. However, that is now what is causing your   issue. The happens, because the user is responsible for adding the security dependency and instantiating required beans as described in the docs. I suggest that you wait for `3.1.5` and then add `spring-security-oauth2-client` and follow the docs for Spring Cloud OpenFeign and Spring Security OAuth2 Client to set it up.

xenoterracide:
I think I did try what's in the docs, any chance there was a problem due to using client credentials and not being in a servlet environment?

OlgaMaciaszek:
@xenoterracide Have you tried against `3.1.5` and set up according to the new recommended version (you will find two approaches in the current docs: one is deprecated - it's indicated in the doc; there's a new one added now as well, so you should use the new approach).

PR #758, opened 2022-09-25
Migrate from Spring Security OAuth to OAuth2 support in Spring Security

dangzhicairang:
#644 Migrate from Spring Security OAuth to OAuth2 support in Spring Security
Open the feature by "spring.cloud.openfeign.oauth2.enabled=true' and the OAuth2AccessTokenInterceptor will inject OAuth2 Token in request header.
By default acquire token by all configured client, if specific the property "spring.cloud.openfeign.oauth2.specifiedClientIds", then it will be used.


OlgaMaciaszek (commented):
Hello @dangzhicairang, thanks for submitting this PR. It has been very helpful.
Setting headers in the Feign `RequestTemplate` looks good. However, I am not exactly convinced by the `findAny` logic regarding the OAuth2 client ids, where the first working `ClientRegistration` from a list of either specified ids or properties is used. I think for security purposes it makes more sense where the user knows exactly which OAuth2 client id they want to use. However, since we can have multiple Feign clients in the application, I'm thinking it might also make sense to retrieve OAuth2 clientIds from Feign Client load-balancing service ids.

dangzhicairang:
> Hello @dangzhicairang, thanks for submitting this PR. It has been very helpful. Setting headers in the Feign `RequestTemplate` looks good. However, I am not exactly convinced by the `findAny` logic regarding the OAuth2 client ids, where the first working `ClientRegistration` from a list of either specified ids or properties is used. I think for security purposes it makes more sense where the user knows exactly which OAuth2 client id they want to use. However, since we can have multiple Feign clients in the application, I'm thinking it might also make sense to retrieve OAuth2 clientIds from Feign Client load-balancing service ids.
> 
> Normally, I would just add this to the PR review. However, since I'd like to get this migration done soon, I have created [a new branch](https://github.com/spring-cloud/spring-cloud-openfeign/commits/migrate-oauth-support-new) based on top of yours (retaining your commits) with those changes. I have also removed the legacy OAuth components and fixed the properties (since the OAuth2 client lib is part of Spring, we do not need to add the dependency to SC OF dependency pom and we should not explicitly specify the version, it will come from the Spring bom).

@OlgaMaciaszek Thanks for your reply, i saw the new branch and understood the logic you said. Which i use the `findAny clientId` logic is for migrate from application that already register some clients, however there are real problems with security, now it looks work well, thanks again.

OlgaMaciaszek:
Sure, thanks for the contribution :)

OlgaMaciaszek:
Closing in favour of https://github.com/spring-cloud/spring-cloud-openfeign/commits/migrate-oauth-support-new.

Issue #756, opened 2022-09-22
Tests fail that require deprecated OAuth2FeignRequestInterceptor


spencergibb:
AccessTokenProviderWithLoadBalancerInterceptorTests and AccessTokenProviderWithoutLoadBalancerInterceptorTests

PR #759, opened 2022-09-27
Migrate oauth support new

OlgaMaciaszek:
Based on https://github.com/spring-cloud/spring-cloud-openfeign/pull/758. See [comment](https://github.com/spring-cloud/spring-cloud-openfeign/pull/758#pullrequestreview-1121963121) for logic discussion.

Fixes gh-644.
Fixes gh-756.


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/OAuth2AccessTokenInterceptor.java
```diff
@@ -0,0 +1,173 @@
...
+			return null;
+		}
+
+		Authentication principal = SecurityContextHolder.getContext().getAuthentication();
+		if (principal == null) {
+			principal = ANONYMOUS_AUTHENTICATION;
+		}
+
+		// Already exist
+		OAuth2AuthorizedClient oAuth2AuthorizedClient = oAuth2AuthorizedClientService.loadAuthorizedClient(clientId,
```
jgrandja, line 143: This block of code (including `notExpired()`) is not needed so can be removed.

The `OAuth2AuthorizedClientManager` will load the `OAuth2AuthorizedClient` (if available) and additionally check if the access token is expired. If expired, it will attempt to re-authorize the access token via one of the `OAuth2AuthorizedClientProvider`.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/OAuth2AccessTokenInterceptor.java
```diff
@@ -0,0 +1,173 @@
...
+				principal.getName());
+		if (oAuth2AuthorizedClient != null) {
+			OAuth2AccessToken accessToken = oAuth2AuthorizedClient.getAccessToken();
+			if (accessToken != null && notExpired(accessToken)) {
+				return accessToken;
+			}
+		}
+
+		OAuth2AuthorizeRequest authorizeRequest = OAuth2AuthorizeRequest.withClientRegistrationId(clientId)
```
jgrandja, line 152: I notice you're using the `clientId` instead of `clientRegistrationId`, which is the expected input. See the difference between `clientId` and `clientRegistrationId` [here](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/core.html#oauth2Client-client-registration)

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/OAuth2AccessTokenInterceptor.java
dangzhicairang: Okay, i saw the logic in `AuthorizedClientServiceOAuth2AuthorizedClientManager` and i will remove this block.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/OAuth2AccessTokenInterceptor.java
dangzhicairang: Yes, i found that `clientRegistrationId` is more suitable,i will adjust it.

dangzhicairang:
@OlgaMaciaszek Based on these reviews, i have done some work on [dangzhicairang:main](https://github.com/dangzhicairang/spring-cloud-openfeign) if you'd like to merge it.

OlgaMaciaszek:
Thanks, @dangzhicairang. Since there are some more changes to be introduced, I will cherry-pick your latest commits onto this branch, then make additional changes and merge it all.

PR #765, opened 2022-10-05
Handle byte[] arrays correctly.(Fixes gh-741)

### LINKED ISSUE #741: The AbstractFormWriter.isTypeOrCollection() cannot handle byte[] arrays
**Assumptions**

There is an endpoint with `byte[]` parameter. E.g. photo parameter in following snippet.
```java
// byte[] parameter
@RequestLine("POST /send_photo")
@Headers("Content-Type: multipart/form-data")
void sendPhoto (@Param("is_public") Boolean isPublic, @Param("photo") byte[] photo);
```
The `sendPhoto()` method is called.

**Issue**

The `AbstractFormWriter.isTypeOrCollection()` tries to cast the `byte[]` value into `Object[]` which leads to the _"java.lang.ClassCastException: class [B cannot be cast to class [Ljava.lang.Object; ([B and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')"_ error.

Problem is on the line 65: https://github.com/spring-cloud/spring-cloud-openfeign/blob/10ad8a0c6ea0a5b641eb31318c93055bdaf4c4f8/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/AbstractFormWriter.java#L63-L65

**Expected behaviour**

The method should be able to handle arrays of primitive types without casting into `Object[]`.

FYI: @darrenfoong

pandaapo:
Fixes #741

Make `AbstractFormWriter#isTypeOrCollection()` handle byte[] arrays correctly instead of casting byte[] to object[] directly.


spencergibb:
Is copying the array the only way around this?

OlgaMaciaszek:
Thanks, @pandaapo . Will review today.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/AbstractFormWriter.java
```diff
@@ -61,10 +62,13 @@ public void write(Output output, String key, Object object) throws EncodeExcepti
...
 	private boolean isTypeOrCollection(Object object, Predicate<Object> isType) {
+		if (object == null) {
+			return false;
+		}
 		if (object.getClass().isArray()) {
-			Object[] array = (Object[]) object;
-
-			return array.length > 1 && isType.test(array[0]);
+			int len = Array.getLength(object);
+			Object one = Array.get(object, 0);
```
OlgaMaciaszek, line 70: We should probably verify first if  the array is not empty.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/AbstractFormWriter.java
```diff
@@ -61,10 +62,13 @@ public void write(Output output, String key, Object object) throws EncodeExcepti
...
+			return len > 1 && one != null && isType.test(one);
```
OlgaMaciaszek, line 71: Shouldn't this be `>= 1`?

OlgaMaciaszek (changes requested):
Thanks @pandaapo . The fix looks mostly good, but have added comments - please address them. Please add your full name with `@author` tag to the javadoc. Could you also add a test that will verify your changes (one that would fail before your change)? Additionally, could you submit these changes against `3.1.x` branch instead of `main` so that they get included in the upcoming `2021.x` releases as well?

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/AbstractFormWriter.java
OlgaMaciaszek: Indeed, it is. In this case, let's not change it for now as there were no issues reported regarding that logic.

OlgaMaciaszek:
Closing in favour of https://github.com/spring-cloud/spring-cloud-openfeign/pull/774.

PR #768, opened 2022-10-11
Remove deprecations and outdated API usage.

OlgaMaciaszek:


docs/src/main/asciidoc/spring-cloud-openfeign.adoc
```diff
@@ -191,7 +191,7 @@ spring:
 						requestInterceptors:
 							- com.example.FooRequestInterceptor
 							- com.example.BarRequestInterceptor
-						decode404: false
+						dismiss404: false
```
spencergibb, line 194: Need to mention this in release notes

OlgaMaciaszek: Will do.

PR #774, opened 2022-10-18
Handle byte[] arrays correctly.(Fixes gh-741)

### LINKED ISSUE #741: The AbstractFormWriter.isTypeOrCollection() cannot handle byte[] arrays

pandaapo:
Fixes #741

Make `AbstractFormWriter#isTypeOrCollection()` handle byte[] arrays correctly instead of casting byte[] to object[] directly.

### DIFF
--- a/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/AbstractFormWriterTest.java
+++ b/spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/AbstractFormWriterTest.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2013-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.openfeign.support;
+
+import java.io.IOException;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.cloud.openfeign.support.AbstractFormWriter;
+import org.springframework.http.MediaType;
+
+/**
+ * @author Wu Daifu
+ */
+public class AbstractFormWriterTest {
+
+	@Test
+	public void testIsApplicable() {
+		MockFormWriter formWriter = new MockFormWriter();
+		Object object = new Object();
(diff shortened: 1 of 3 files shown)


spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/AbstractFormWriterTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.cloud.openfeign.support.AbstractFormWriter;
+import org.springframework.http.MediaType;
+
+/**
+ * @author Wu Daifu
+ */
+public class AbstractFormWriterTest {
```
OlgaMaciaszek, line 30: Please rename the test to `AbstractFormWriterTests`. Also, please remove the unnecessary `public` keyword from the class and method signatures.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/AbstractFormWriterTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+	@Test
+	public void testIsApplicable() {
```
OlgaMaciaszek, line 33: Please rename the method to something more meaningful like `shouldCorrectlyResolveIfApplicableForCollection`.

OlgaMaciaszek (changes requested):
Thanks, @pandaapo. Just some cosmetic changes are required for the test. Please see the new comments.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/AbstractFormWriterTest.java
leopardrota: OK. Should I modify this on the PR of `main` branch as well?

OlgaMaciaszek (changes requested):
Thanks, @pandaapo. There's a checkstyle issue (see build results). Please fix.

PR #775, opened 2022-10-19
Release RequestAttributes when called asynchronously. Fixes gh-716

### LINKED ISSUE #716: Request attributes exposed by circuit breakers are preserved permanently when called asynchronously
As shown in the code below, the request attributes are passed to the thread calling the execute method through the `RequestContextHolder` (ThreadLocal)

https://github.com/spring-cloud/spring-cloud-openfeign/blob/8493de9aa6038024acab81142c5cb714d1b9edfc/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java#L124-L136

This code is from: https://github.com/spring-cloud/spring-cloud-openfeign/issues/193
And updated: https://github.com/spring-cloud/spring-cloud-openfeign/issues/572

This creates a conflict. The idea here should be to use ThreadLocal to pass request attributes in async calls and do nothing in sync calls. And now ThreadLocal is not cleaned up after using it in async call.

The easiest way is to compare whether the calling thread and the executing thread are consistent before the operation:
```java
final RequestAttributes requestAttributes = RequestContextHolder.getRequestAttributes();
final Thread caller = Thread.currentThread();
return () -> {
                boolean isAsync = caller != Thread.currentThread();
                try {
                    if (isAsync) {
                        RequestContextHolder.setRequestAttributes(requestAttributes);
                    }
                    return dispatch.get(method).invoke(args);
                }
                catch (RuntimeException throwable) {
                    throw throwable;
                }
                catch (Throwable throwable) {
                    throw new RuntimeException(throwable);
                } finally {
                    if (isAsync) {
                        RequestContextHolder.resetRequestAttributes();
                    }
                }
 };
```

vicasong:
Fixes #716

Make `FeignCircuitBreakerInvocationHandler#asSupplier` compare the calling thread and the executing thread for consistency before the operation, use `ThreadLocal` to pass request attributes in the asynchronous call, and do nothing in the synchronous call.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignCircuitBreakerInvocationHandler.java
@@ -36,6 +36,14 @@ import org.springframework.web.context.request.RequestContextHolder;
 
 import static feign.Util.checkNotNull;
 
+/**
+ * @author Grzejszczak
+ * @author Sharma
+ * @author Niang
+ * @author Bohutskyi
+ * @author kim
+ * @author Vicasong
+ */
 class FeignCircuitBreakerInvocationHandler implements InvocationHandler {
 
 	private final CircuitBreakerFactory factory;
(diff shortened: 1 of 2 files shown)


OlgaMaciaszek:
Fixes gh-716.

OlgaMaciaszek (approved):
Thanks @vicasong. LGTM.

PR #776, opened 2022-10-19
refactor load balance client so that users can add their custom implementation

liubao68:
>>> Why add this refactor ?
   
   My application need to add a custom Client like FeignBlockingLoadBalancerClient, e.g.

 ```
public class CustomFeignBlockingLoadBalancerClient implements Client {
   ...
}
```

This works pretty well when using load balanced clients. But when mix with load balanced and non load balanced clients, `FeignClientFactoryBean` hard coded to work with FeignBlockingLoadBalancerClient to get the delegate

```
		if (client != null) {
			if (client instanceof FeignBlockingLoadBalancerClient) {
				// not load balancing because we have a url,
				// but Spring Cloud LoadBalancer is on the classpath, so unwrap
				client = ((FeignBlockingLoadBalancerClient) client).getDelegate();
			}
			if (client instanceof RetryableFeignBlockingLoadBalancerClient) {
				// not load balancing because we have a url,
				// but Spring Cloud LoadBalancer is on the classpath, so unwrap
				client = ((RetryableFeignBlockingLoadBalancerClient) client).getDelegate();
			}
			builder.client(client);
		}
```

Refactor the code can help easier to customize FeignBlockingLoadBalancerClient and work with  non load balanced clients.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
@@ -48,6 +48,7 @@ import org.springframework.beans.factory.FactoryBean;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.cloud.openfeign.clientconfig.FeignClientConfigurer;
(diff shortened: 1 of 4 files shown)


spencergibb:
We will not do this at all on the 3.1.x line

liubao68:
> We will not do this at all on the 3.1.x line

Thanks for your reply, this PR can be a reference for a future improvement for the new version. I do not have a latest version compile environment yet, and I just create a PR for 3.1.x.

PR #789, opened 2022-11-15
Remove additional Commons HttpClient dependencies.  

ryanjbaxter:
Additional Fix For #788


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/clientconfig/OkHttpFeignConfiguration.java
```diff
@@ -1,77 +0,0 @@
-/*
```
ryanjbaxter, line 1: This is the main thing I was unsure about.  I did not understand why we had this configuration class.  It pretty much duplicated what was in `FeignAutoConfiguration` and was only used in `OkHttpFeignLoadBalancerConfiguration`.  Seems to me we can just remove this and use whats in `FeignAutoConfiguration`

PR #462, opened 2021-01-18
Micrometer Support

jonatan-ivanov:
Draft PR to start discussion about #457. 
The scope of the PR is a little wider than just setting up micrometer for Feign, it makes spring-cloud-openfeign capable of configuring any [Capability](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/Capability.java) (e.g.: [Metrics5Capability](https://github.com/OpenFeign/feign/blob/master/dropwizard-metrics5/src/main/java/feign/metrics5/Metrics5Capability.java) for Dropwizard Metrics or [MicrometerCapability](https://github.com/OpenFeign/feign/blob/master/micrometer/src/main/java/feign/micrometer/MicrometerCapability.java) for micrometer).

It also autoconfigures [MicrometerCapability](https://github.com/OpenFeign/feign/blob/master/micrometer/src/main/java/feign/micrometer/MicrometerCapability.java) if a `MeterRegistry` is configured and `feign-micrometer` is on the classpath.

Also, this PR is not complete, a few things are missing, I'm going to fix/add them after we agreed on the details below this section:
- [x] Fixing current tests (some of them are failing)
- [x] Adding new tests
- [x] Adding docs
- [x] Adding `@ConfigurationProperties` for the enabled flag

Topics to discuss:
- What do you think about the changes in general?
- On what level should the enabled flag be (client level can be more complicated)?
- CheckStyle messes up the formatting (see: Stream API, it should be one op/line)
- Fyi: capabilities are sorted right now but they are decorators so they should be

PR #793, opened 2022-11-19
Micrometer Observation API Support

jonatan-ivanov:
This change auto-configures `MicrometerObservationCapability`. The previous auto-configuration for `MicrometerCapability` is still available but for that `MicrometerObservationCapability` should not be present (disabled, no observation registry, micrometer-observation is not on the classpath).

**Breaking change:** properties were renamed from `spring.cloud.openfeign.metrics` to `spring.cloud.openfeign.micrometer`.

Previous Micrometer support in SC-OpenFeign: https://github.com/spring-cloud/spring-cloud-openfeign/pull/462
`MicrometerObservationCapability` support in Feign: https://github.com/OpenFeign/feign/pull/1760

### DIFF
--- a/spring-cloud-openfeign-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-cloud-openfeign-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -57,9 +57,9 @@
 			"defaultValue": "false"
 		},
 		{
-			"name": "spring.cloud.openfeign.metrics.enabled",
+			"name": "spring.cloud.openfeign.micrometer.enabled",
 			"type": "java.lang.Boolean",
-			"description": "Enables metrics capability for Feign.",
+			"description": "Enables Micrometer capabilities for Feign.",
 			"defaultValue": "true"
 		},
 		{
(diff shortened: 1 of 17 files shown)


spring-cloud-openfeign-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -57,9 +57,9 @@
 			"defaultValue": "false"
 		},
 		{
-			"name": "spring.cloud.openfeign.metrics.enabled",
+			"name": "spring.cloud.openfeign.micrometer.enabled",
```
spencergibb, line 60: Does the property name need to change?

jonatan-ivanov: I think yes. I don't think `metrics.enabled` makes any sense when the users don't even have metrics, e.g.: they enable tracing but not metrics through the Observation API.
Or to put it in a different way: SC-OpenFeign has no idea what the user will configure in their `ObservationRegistry`.

marcingrzejszczak (approved):

OlgaMaciaszek (approved):
Thanks a lot @jonatan-ivanov. LGTM.

Issue #625, opened 2021-11-10
For hc5 hostname verification disable-ssl-validation does not turn off hostname verification


bertoltmeier:
When using feignclient with option disable-ssl-validation: true and hc5 enabled, hostnameverification is not turned off like it is when using httpclient4.

We use option disable-ssl-validation often for local development or testing. Is it possible to deactivate hostname verification for feignclient/hc5 by configuration somehow?

My configuration (application.yml):

```
feign:
  httpclient:
    disable-ssl-validation: true
    hc5:
      enabled: true
```

Exception I retrieve:

```
2021-11-10 23:13:25.968 DEBUG 91492 --- [           main] u.c.RestControllerApiClient : [RestControllerApiClient#retrieveSomethingUsingPOST] ---> END HTTP (178-byte body)
2021-11-10 23:13:26.133 DEBUG 91492 --- [           main] u.c.RestControllerApiClient : [RestControllerApiClient#retrieveSomethingUsingPOST] <--- ERROR SSLHandshakeException: No name matching localhost found (164ms)
2021-11-10 23:13:26.134 DEBUG 91492 --- [           main] u.c.RestControllerApiClient : [RestControllerApiClient#retrieveSomethingUsingPOST] javax.net.ssl.SSLHandshakeException: No name matching localhost found
	at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:131)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:349)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:292)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:287)
	at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.checkServerCerts(CertificateMessage.java:654)
	at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.onCertificate(CertificateMessage.java:473)
	at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.consume(CertificateMessage.java:369)
	at java.base/sun.security.ssl.SSLHandshake.consume(SSLHandshake.java:392)
	at java.base/sun.security.ssl.HandshakeContext.dispatch(HandshakeContext.java:443)
	at java.base/sun.security.ssl.HandshakeContext.dispatch(HandshakeContext.java:421)
	at java.base/sun.security.ssl.TransportContext.dispatch(TransportContext.java:182)
[...]

PR #798, opened 2022-12-06
Disable hostname verification when ssl validation is disabled.

emopti-jrufer:
Fixes gh-625.

Prior to Apache Http Client5 `org.springframework.cloud.commons.httpclient.DefaultApacheHttpClientConnectionManagerFactory` was being used which added the `NoopHostnameVerifier`.  See https://github.com/spring-cloud/spring-cloud-commons/blob/e72fc6c6906cab8dedd1277dccffb5aeca99aec6/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/httpclient/DefaultApacheHttpClientConnectionManagerFactory.java#L69


emopti-cmeyer:
@ryanjbaxter @OlgaMaciaszek hoping you can take a look and get this merged before release to avoid unnecessary breakage. This is a 1-line fix for an issue that was introduced by the switch from Apache HttpClient 4 to 5.

@ryanjbaxter committed similar code for OkHttpClient a few weeks ago here: https://github.com/spring-cloud/spring-cloud-openfeign/commit/1ec3c6522d563401656ee8b9338a528388381023

All this pull request does is to make sure that when the `spring.cloud.openfeign.httpclient.disable-ssl-validation` property is set to `true`, the HttpClient also ignores host name mismatches (e.g. the common name from the certificate doesn't have to match).

As described in the original pull request description, the Apache HttpClient 4 implementation was also using the NoopHostnameVerifier and as mentioned in my comment here w/ link, the OkHttpClient does the same.

PR #880, opened 2023-07-27
Remove Codacy badge

AlejandroDelPapa:
It is not working anymore, maybe the URL changed, maybe the token is invalid


OlgaMaciaszek:
Thanks for the PR @AlejandroDelPapa. We've handled it while switching the docs to Antora.

PR #92, opened 2018-12-07
issues 81

hsoftxl:
https://github.com/spring-cloud/spring-cloud-openfeign/issues/81

PR #924, opened 2023-11-05
add tests when the `FeignClient.path` is set

AlexElin:
Add tests when the FeignClient.path is set to demonstrate the issue https://github.com/spring-cloud/spring-cloud-openfeign/issues/92. One of them fails.


OlgaMaciaszek:
Merged within 502613bb740957e9d96feec99077c6ccc0e75351.

Issue #923, opened 2023-11-05
Incorrect resolved url when the `FeignClient.path` is set and the base url is defined in the properties file


AlexElin:
**Describe the bug**
When the `FeignClient.path` is set and a url is defined in the properties file, then the incorrect url is constructed. So the call fails with an exception

**Steps to reproduce**
1. Set up the target host in the properties file
```
spring.cloud.openfeign.client.config.withPathAndUrlFromConfigClient.url=http://localhost:7777
```
2. Set up the `FeignClient.path` (which is a common prefix according to the its javadoc)
```java
@FeignClient(name = "withPathAndUrlFromConfigClient", path = "/common")
interface WithPathAndUrlFromConfigClient {

	@GetMapping("/test")
	UrlTestClient.UrlResponseForTests test();

}
```
3. Call the `test()` method.

**Actual result**: the exception is thrown
```
java.lang.IllegalArgumentException: target values must be absolute.

	at feign.RequestTemplate.target(RequestTemplate.java:504)
	at feign.Target$HardCodedTarget.apply(Target.java:101)
	at feign.SynchronousMethodHandler.targetRequest(SynchronousMethodHandler.java:126)
	at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:91)
	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:70)
	at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:96)
	at org.springframework.cloud.openfeign.$Proxy187.test(Unknown Source)
	at org.springframework.cloud.openfeign.NonRefreshableFeignClientUrlTests.shouldInstantiateFeignClientWhenUrlFromPropertiesAndPathInTheFeignClientAnnotation(NonRefreshableFeignClientUrlTests.java:93)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```
**Expected result**: the url `http://localhost:7777/common/test` is called.

**Note**: when I hardcode the url into the `FeignClient.url` then it works fine.

Spring Cloud Open Feign version: 4.0.4

**Sample**
https://github.com/spring-cloud/spring-cloud-openfeign/pull/924 demonstrates the issue

OlgaMaciaszek:
Hi @AlexElin - thanks for reporting the issue.

OlgaMaciaszek:
Closed with commits: 
[...]

PR #927, opened 2023-11-09
(fix): Incorrect resolved url when the FeignClient.path is set and the base url is defined in the properties file

AlexElin:
Main changes:
1. Decorate a target with new `PathPrefixedTarget` target that adds path to the url.
2. Change `Targeter.target`'s signature: Target.HardcodedTarget -> Target
3. Cleanup tests from things which should not be checked

Closes https://github.com/spring-cloud/spring-cloud-openfeign/issues/923, https://github.com/spring-cloud/spring-cloud-openfeign/pull/924


OlgaMaciaszek (commented):
Thanks for submitting the PR, @AlexElin. I think I'd prefer to go with the more succinct solution proposed by @cbezmen in gh-939. However, I'd like your tests from gh-924 to be cherry-picked into the final solution, since you've originally proposed them.

PR #939, opened 2023-11-27
Fix issue 923 config url and feign annotation path

cbezmen:
Related issue -> https://github.com/spring-cloud/spring-cloud-openfeign/issues/923


OlgaMaciaszek (changes requested):
Thanks for the PR @cbezmen. In general looks good, but I've added a few comments. Also, could you please submit it against `4.0.x` instead of `main`, so that we get the fix also into the `2022.0.x` release train?

OlgaMaciaszek:
Fixes gh-923.

OlgaMaciaszek:
Merged within 44657956422bd30deeb4881bf73901a7b906894a.

cbezmen:
Thanks @OlgaMaciaszek

OlgaMaciaszek:
Thanks for working on the issue @cbezmen.

PR #971, opened 2024-01-26
interface-annotations-doc

galaxy-sea:
I often feel confused about which annotations can be used, and I hope this PR can also help everyone


OlgaMaciaszek:
Have looked at this now, but these elements are not actually semantically same - they behave differently, take in different arguments, etc.. It would have to be described in much more detail to be accurate. Also, SC OF is an integration on top of Feign. If someone would like to use Feign's own annotations, they should check the documentation for the core Feign project.

PR #995, opened 2024-02-26
New Feature 

harishwar-babu:
Added new Feature JWT-Access-Token-Interceptor
* In the previous versions. developers need to manually implement the requestInterceptor. to pass the token between services.
* By including this feature. Developer just need to create a bean, for this class and there by the implementation is done by default.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/JwtAccessTokenInterceptor.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/JwtAccessTokenInterceptor.java
@@ -1,0 +1,33 @@
+package org.springframework.cloud.openfeign.security;
+
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


OlgaMaciaszek:
Hello, @harishwar-babu, a separate `HttpServletRequest` object is created for every HTTP request. How would you inject it as a field during bean creation?

harishwar-babu:
As HttpServletRequest is already a component. So whenever the developer
tries to use my Feature called JwtAccessTokenInterceptor, they need to
create a bean for my feature upon creation, this httpservletRequest is also
being created and injected by using the field injection strategy.

Thanks
Harish Babu

OlgaMaciaszek:
Thanks, @harishwar-babu, but we are not going to be adding this feature at this point.

harishwar-babu:
May I know the reason?

Thanks

OlgaMaciaszek:
`HttpServletRequest` objects correspond to specific requests, while beans should correspond to components present throughout the lifecycle of the application.

PR #1003, opened 2024-03-15
optimize ResponseInterceptor

galaxy-sea:
optimize ResponseInterceptor

### DIFF
--- a/docs/modules/ROOT/pages/spring-cloud-openfeign.adoc
+++ b/docs/modules/ROOT/pages/spring-cloud-openfeign.adoc
@@ -204,7 +204,9 @@ spring:
 						requestInterceptors:
 							- com.example.FooRequestInterceptor
 							- com.example.BarRequestInterceptor
-						responseInterceptor: com.example.BazResponseInterceptor
+						responseInterceptors:
+							- com.example.BazResponseInterceptor
+							- com.example.QuxResponseInterceptor
 						dismiss404: false
 						encoder: com.example.SimpleEncoder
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/spring-cloud-openfeign.adoc
```diff
@@ -204,7 +204,9 @@ spring:
 						requestInterceptors:
 							- com.example.FooRequestInterceptor
 							- com.example.BarRequestInterceptor
-						responseInterceptor: com.example.BazResponseInterceptor
+						responseInterceptors:
```
spencergibb, line 207: This is a breaking change unfortunately

spencergibb (changes requested):

docs/modules/ROOT/pages/spring-cloud-openfeign.adoc
galaxy-sea: hello @spencergibb,  Can compatibility be achieved?  Keep `responseInterceptor` and add `responseInterceptors`

OlgaMaciaszek: @galaxy-sea  It could be done, but seems rather messy in terms of the API. I think we will consider it for a major release if it's scheduled.

galaxy-sea: @OlgaMaciaszek , I agree with your opinion

PR #1019, opened 2024-04-09
New Feature

harishwar-babu:
Added new Feature JWT-Access-Token-Interceptor

In the previous versions. developers need to manually implement the requestInterceptor. to pass the token between services.
By including this feature. Developer just need to create a bean, for this class and there by the implementation is done by default.


OlgaMaciaszek:
Hello @harishwar-babu. Thanks for creating the PR. Each HTTP request will have a corresponding `HttpServletRequest` , so this is not somehting that should be injected into the interceptor as a field.

OlgaMaciaszek:
Closing as not planned.

PR #1022, opened 2024-04-17
Ignore unresolved property placeholders in `url`

elkkhan:
fixes https://github.com/spring-cloud/spring-cloud-openfeign/issues/1023

`FeignClientRegistar` seems to [purposefully ignore](https://github.com/spring-cloud/spring-cloud-openfeign/blob/ca4de4e703a953504c3bb7b08fa8c69566697953/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java#L116) unresolved URLs that are SpEL but doesn't do so for property placeholders like `${placeholder}`.
For ignored SpEL strings, the input string is returned, but for placeholders, it tries to [parse it into a `URL` object](https://github.com/spring-cloud/spring-cloud-openfeign/blob/ca4de4e703a953504c3bb7b08fa8c69566697953/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java#L121).
Up until JDK 20, this would be OK and would not fail - and unresolved placeholder URL would probably fail somewhere later down the flow. This is because the underlying implementation of the `URL` class doesn't actually try to parse the host until unless the `URL.openConnection()` is called, which is not called in this method.

[Since JDK 20](https://bugs.openjdk.org/browse/JDK-8293590), this behaviour has changed and the underlying implementation of the `URL` class now parses the host eagerly without waiting for a `URL.openConnection()` call.

This was discovered when our tests that use [@JsonTest](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/autoconfigure/json/JsonTest.html) started failing once we upgraded to JDK 20+

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java
@@ -80,6 +80,14 @@ class FeignClientsRegistrar implements ImportBeanDefinitionRegistrar, ResourceLo
 	FeignClientsRegistrar() {
 	}
 
+	private static boolean isSpEL(String value) {
+		return value.startsWith("#{") && value.contains("}");
+	}
+
+	private static boolean isPlaceholder(String value) {
+		return value.startsWith("${") && value.endsWith("}");
+	}
+
 	static void validateFallback(final Class clazz) {
 		Assert.isTrue(!clazz.isInterface(), "Fallback class must implement the interface annotated by @FeignClient");
 	}
@@ -113,7 +121,7 @@ class FeignClientsRegistrar implements ImportBeanDefinitionRegistrar, ResourceLo
 	}
 
 	static String getUrl(String url) {
-		if (StringUtils.hasText(url) && !(url.startsWith("#{") && url.contains("}"))) {
+		if (StringUtils.hasText(url) && !isSpEL(url) && !isPlaceholder(url)) {
 			if (!url.contains("://")) {
 				url = "http://" + url;
 			}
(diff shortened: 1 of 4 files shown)


spencergibb:
What exception is thrown if the placeholder is not ignored?

spencergibb:
It doesn't ignore them, it just doesn't prepend `http://`

elkkhan:
hi @spencergibb, 
> It doesn't ignore them, it just doesn't prepend http://

Yes, what I meant by `ignoring` is just that it early returns the input value instead of letting it go through below. so as a consequence it throws an `IllegalArgumentException` upon catching a `MalformedURLException`

```java
if (!url.contains("://")) {
    url = "http://" + url;
}
try {
    new URL(url);
}
catch (MalformedURLException e) {
    throw new IllegalArgumentException(url + " is malformed", e);
}
```

spencergibb:
Why isn't that preferred, to error early rather than later?

elkkhan:
> Why isn't that preferred, to error early rather than later?

I actually would prefer that too but I would expect the behaviour to be consistent between SpEL and placeholders.
I chose to let placeholders through to avoid breaking existing code which would be a greater risk if we also let SpEL fail early. I'm not too sure about the volumes here though, as in how realistic/often is it for unresolved URLs to get to that method in the first place

elkkhan:
👋 @spencergibb!
any thoughts on above?

elkkhan:
hey @spencergibb @OlgaMaciaszek,
just checking to see if you have any further thoughts on this

OlgaMaciaszek:
@elkkhan Thanks for creating this PR. I agree with @spencergibb that it would be preferable to get an early exception. At the same time, I agree with you that consistency is important. However, we should unify it towards the more desirable outcome. Given the historical handling of SpEl, we cannot introduce this change now (in a non-major release), but you could create a PR and we'd mark it as one for a major release and merge it whenever we start working on one. Feel free to create it. I'm going to close this one.

PR #1027, opened 2024-04-20
Fix Reactive Support doc

Puppy4C:


OlgaMaciaszek (approved):
Thanks for the fix @Puppy4C. LGTM.

PR #1038, opened 2024-06-21
polishing FeignClientsRegistrar

birariro:
```getName (Map<String, Object>)``` does not need to be for test purposes.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2022 the original author or authors.
+ * Copyright 2013-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


OlgaMaciaszek (commented):
Hello @birariro, I'm not sure why you've proposed this change. We do need to use `beanFactory` to properly resolve the attributes.

birariro:
Hello @OlgaMaciaszek 

```String getName(ConfigurableBeanFactory beanFactory, Map<String, Object> attributes)```
Because the only way to call is ```String getName(Map<String, Object> attributes)```
Isn't beanFactory always null?

And 
```/* for testing */ String getName(Map<String, Object> attributes) ``` 
If you look at the comments, it says it's for testing, but if it's unnecessary, it looked better without it

spencergibb:
Keep the for testing comment

OlgaMaciaszek:
@birariro  It's helpful to know that the visibility level has been raised for testing purposes. That's why we put these comments there. I thought `getName` was also called at a different point, but possibly after some refactoring it isn't, so right, we can change that.

OlgaMaciaszek (changes requested):
Thanks @birariro. Looks good. Can you please just add your full name and surname with `@author` tag to the javadoc of the class and update the date in the license comment to `2013-2024`?

birariro:
@OlgaMaciaszek I have applied the changes as requested.

PR #1041, opened 2024-07-07
fix: defaultValue in additional-spring-configuration-metadata.json

dowenliu-xyz:
defaultValue of `spring.cloud.openfeign.circuitbreaker.alphanumeric-ids.enabled` is `true` since version 4.0.0


OlgaMaciaszek (changes requested):
Hi @dowenliu-xyz. Thanks for submitting the PR. Looks good, however, could you please submit it against `4.1.x` instead of `main`? Like this, the change will also be included in the `2023.0.x` release train.

dowenliu-xyz:
Ok, it's fine. And I've noticed that I have make this change base on main branch. Is it ok?

OlgaMaciaszek:
> Ok, it's fine. And I've noticed that I have make this change base on main branch. Is it ok?

Yes, looks fine.

OlgaMaciaszek (approved):
LGTM. Thanks.

PR #1050, opened 2024-08-19
Authorize with anonymous principal if principal name is null

### LINKED ISSUE #1049: OAuth2AccessTokenInterceptor can handle Authentication Principal where principalName is null
When `OAuth2AccessTokenInterceptor` encounters an Authentication Principal where `getName` is `null`, it passes the `null` value on to the `OAuth2AuthorizedClientManager`, which in practice usually results in a runtime exception like `principalName cannot be empty`.

The code already checks for a null principal an in that case passes ANONYMOUS_AUTHENTICATION. It appears to also make sense to pass ANONYMOUS_AUTHENTICATION for cases with a non-null principal but a null principal name.

Handling a null principal name downstream or upstream appears more complicated and it feels more natural to follow the current implementation of employing the ANONYMOUS_AUTHENTICATION placeholder for cases where the current authentication principal is not suitable.

**Additional context**
None

agileknight:
The authorizedClientManager.authorize method requires a non-null principal name or it will usually throw an exception in practice like "principalName cannot be empty".

Using the anonymous principal in this case like for a null principal handles the situation more gracefully.

Fixes #1049

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/OAuth2AccessTokenInterceptor.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/security/OAuth2AccessTokenInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2022 the original author or authors.
+ * Copyright 2015-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


OlgaMaciaszek (changes requested):
Hi @agileknight. Thanks for submitting the PR. In general, looks good. Please add your name with `@author` tag to the javadocs of all the files you've modified. Please update the date in the license comment in the files you've modified to `2015-2024`. 
Also, please submit this PR against `4.1.x` instead, so that it also gets included in the `2023.0.x` release train.

OlgaMaciaszek (changes requested):
Some build tests are failing. Please take  a look.

OlgaMaciaszek (commented):
Thanks, @agileknight. LGTM.

Issue #1048, opened 2024-08-16
Misleading Content-Encoding Header in Default-Setup for Request Compression


AndreTeigler:
**Describe the bug**
When the openfeign request compression is used the the `FeignContentGzipEncodingInterceptor` adds two values to the header:

`addHeader(template, HttpEncoding.CONTENT_ENCODING_HEADER, HttpEncoding.GZIP_ENCODING, HttpEncoding.DEFLATE_ENCODING);`

The compression itself is dependent on the used client. The feign default client (https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/Client.java) can either use gzip compression or deflate compression. If both headers are present it will currently choose gzip but the headers remain unchanged.

According to the Content-Encoding documentation (https://datatracker.ietf.org/doc/html/rfc2616#section-14.11) the header indicates that both compressions have been applied. The default client only applies one so the receiving server (if it handles the Content-Encoding header correctly) will first try to decompress with defalte which fails.

**Sample**
Can be reproduced by using a simple spring boot app using spring-cloud-starter-openfeign and setting up a client which will sent a body with request compression. The headers will contain `Content-Enconding: gzip, deflate` but the compressed body will only be gzip data.

OlgaMaciaszek:
Thanks @AndreTeigler, makes sense; we can't  know which client and which setup will be used in the app configuration, however, we should allow the users to apply a more fine-grained configuration via properties.

AndreTeigler:
@OlgaMaciaszek
I could provide a pull request over the next weekend by just adding another configuration property for the encoding header. It defaults to 'gzip, deflate' to keep backwards compatibility but would allow to set the header to the appropiate value for the setup in use.

OlgaMaciaszek:
Thanks @AndreTeigler. Sounds good.

PR #1071, opened 2024-09-08
Make Content-Encoding configurable

AndreTeigler:
A new configuration property was added to make the Content-Encoding configurable when using the feign request compression. The default values are the same as the previous hardcoded values.

Fixes gh-1048


spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isFalse();
+	}
+
+	@Test
+	void apply_SizeEqualsThreshold_NoCompressionHeader() {
```
OlgaMaciaszek, line 51: Please rename to `shouldNotAddCompressionHeaderWhenSizeEqualsThreshold`

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isFalse();
+	}
+
+	@Test
+	void apply_SizeAboveThreshold_DefaultCompressionHeader() {
```
OlgaMaciaszek, line 68: Please rename to `shouldAddCompressionHeaderWhenSizeAboveThreshold`

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+		final var template = new RequestTemplate();
+
+		template.header(HttpEncoding.CONTENT_LENGTH, "2047");
+		template.header(HttpEncoding.CONTENT_TYPE, "application/json");
+
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isFalse();
```
OlgaMaciaszek, line 47: Change to `assertThat(template.headers()).doesNotContainKey(HttpEncoding.CONTENT_ENCODING_HEADER);`

OlgaMaciaszek: Also please change in other tests when more specific assertions can be used instead of `isFalse` or `isTrue`.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isTrue();
+		assertThat(template.headers().get(HttpEncoding.CONTENT_ENCODING_HEADER)).isEqualTo(List.of("gzip", "deflate"));
+	}
+
+	@Test
+	void apply_MimeTypeMismatch_NoCompressionHeader() {
```
OlgaMaciaszek, line 86: Please rename to `shouldNotAddCompressionHeaderWhenTypeMismatch`.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isFalse();
+	}
+
+	@Test
+	void apply_NoMimeTypeSet_DefaultCompressionHeader() {
```
OlgaMaciaszek, line 104: Please rename to `shouldAddDefaultCompressionHeaderWhenMimeTypeNotSet`.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isTrue();
+		assertThat(template.headers().get(HttpEncoding.CONTENT_ENCODING_HEADER)).isEqualTo(List.of("gzip", "deflate"));
+	}
+
+	@Test
+	void apply_NullMimeTypeSet_DefaultCompressionHeader() {
```
OlgaMaciaszek, line 123: Please rename to `shouldAddDefaultCompressionHeaderWhenNullMimeTypeSet`.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isTrue();
+		assertThat(template.headers().get(HttpEncoding.CONTENT_ENCODING_HEADER)).isEqualTo(List.of("gzip", "deflate"));
+	}
+
+	@Test
+	void apply_NonDefaultContentEncodingSet_CustomCompressionHeader() {
```
OlgaMaciaszek, line 142: Please rename to `shouldAddCustomCompressionHeader`.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+		final var interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isTrue();
+		assertThat(template.headers().get(HttpEncoding.CONTENT_ENCODING_HEADER)).isEqualTo(List.of("gzip"));
+	}
+
+	@Test
+	void apply_NoContentEncodingsSet_DefaultCompressionHeader() {
```
OlgaMaciaszek, line 161: This test will need changing after behaviour change.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTest.java
```diff
@@ -0,0 +1,198 @@
...
+		interceptor.apply(template);
+
+		assertThat(template.headers().containsKey(HttpEncoding.CONTENT_ENCODING_HEADER)).isTrue();
+		assertThat(template.headers().get(HttpEncoding.CONTENT_ENCODING_HEADER)).isEqualTo(List.of("gzip", "deflate"));
+	}
+
+	@Test
+	void apply_NullContentEncodingsSet_DefaultCompressionHeader() {
+
+		final var properties = new FeignClientEncodingProperties();
```
OlgaMaciaszek, line 182: This test will need changing after behaviour change.

OlgaMaciaszek (changes requested):
Hi @AndreTeigler, thanks for the PR. Have requested some changes, mostly minor. Please address the comments.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptor.java
```diff
@@ -44,11 +44,18 @@ protected FeignContentGzipEncodingInterceptor(FeignClientEncodingProperties prop
...
 
 		if (requiresCompression(template)) {
-			addHeader(template, HttpEncoding.CONTENT_ENCODING_HEADER, HttpEncoding.GZIP_ENCODING,
-					HttpEncoding.DEFLATE_ENCODING);
+			addHeader(template, HttpEncoding.CONTENT_ENCODING_HEADER, getContentEncodings());
 		}
 	}
 
+	private String[] getContentEncodings() {
+		if (getProperties().getContentEncodings() != null && getProperties().getContentEncodings().length > 0) {
```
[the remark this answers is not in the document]
AndreTeigler: I wanted to stay inline with the code in 'matchesMimeType' but I see the point that this is too silent. I just fallback to the default without informing the user that something went wrong.

What about logging a warning and then use the fallback value? Would still be inline with the property evaluation behavior in this class but would give the user a chance to see the misconfiguration?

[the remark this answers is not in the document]
AndreTeigler: I added the log as suggestion, but I can also throw an exception if this is more appropiate.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptor.java
```diff
@@ -44,11 +48,19 @@ protected FeignContentGzipEncodingInterceptor(FeignClientEncodingProperties prop
...
+			addHeader(template, HttpEncoding.CONTENT_ENCODING_HEADER, getContentEncodings());
 		}
 	}
 
+	private String[] getContentEncodings() {
+		if (getProperties().getContentEncodingTypes() != null && getProperties().getContentEncodingTypes().length > 0) {
+			return getProperties().getContentEncodingTypes();
+		}
+
+		log.warn("Invalid content encoding configuration, falling back to default.");
```
OlgaMaciaszek, line 60: Many users may ignore warnings and if  they changed something in that prop, it means they expected something other than the default prop value to appear there. I think it's better if they know something went wrong, so I think we should throw something like an `IllegalStateException` instead.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTests.java
```diff
@@ -0,0 +1,181 @@
...
+
+import java.util.List;
+
+import feign.RequestTemplate;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for {@link FeignContentGzipEncodingInterceptor}
```
OlgaMaciaszek, line 27: Please add a full stop at the end of this line.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTests.java
```diff
@@ -0,0 +1,181 @@
...
+		interceptor.apply(template);
+
+		assertThat(template.headers()).doesNotContainKey(HttpEncoding.CONTENT_ENCODING_HEADER);
+	}
+
+	@Test
+	void shouldAddDefaultCompressionHeaderWhenMimeTypeNotSet() {
+		final FeignClientEncodingProperties properties = new FeignClientEncodingProperties();
+		properties.setMimeTypes(new String[] {});
+
```
OlgaMaciaszek, line 99: Please remove this empty line, so that the test only has 2 sections divided by empty lines: Arrange: all preliminary setup, A

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTests.java
```diff
@@ -0,0 +1,181 @@
...
+		final FeignContentGzipEncodingInterceptor interceptor = new FeignContentGzipEncodingInterceptor(properties);
+
+		interceptor.apply(template);
+
+		assertThat(template.headers()).containsKey(HttpEncoding.CONTENT_ENCODING_HEADER);
+		assertThat(template.headers().get(HttpEncoding.CONTENT_ENCODING_HEADER)).isEqualTo(List.of("gzip", "deflate"));
+	}
+
+	@Test
+	void shouldAddDefaultCompressionHeaderWhenNullContentEncodingTypes() {
```
OlgaMaciaszek, line 165: Test should be modified after implementation is adjusted.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptorTests.java
```diff
@@ -0,0 +1,181 @@
...
+
+		interceptor.apply(template);
+
+		assertThat(template.headers()).containsKey(HttpEncoding.CONTENT_ENCODING_HEADER);
+		assertThat(template.headers().get(HttpEncoding.CONTENT_ENCODING_HEADER)).isEqualTo(List.of("gzip"));
+	}
+
+	@Test
+	void shouldAddDefaultCompressionHeaderWhenContentEncodingTypesAreEmpty() {
+		final FeignClientEncodingProperties properties = new FeignClientEncodingProperties();
```
OlgaMaciaszek, line 149: Test should be modified after implementation is adjusted.

OlgaMaciaszek (changes requested):
Thanks a lot @AndreTeigler. I have added some more comments.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/encoding/FeignContentGzipEncodingInterceptor.java
AndreTeigler: I try to read warnings in my apps :).

OlgaMaciaszek: So do many users, but if we know that the user has explicitly opted out of the default behaviour by changing sth in the properties file, they should get an exception rather than the default behaviour they do not expect, as it's always better to know that something hasn't gone as expected.

OlgaMaciaszek (approved):
Thanks @AndreTeigler. LGTM.

Issue #1098, opened 2024-10-09
spring-cloud-openfeign-core transitively depends on commons-io with CVE-2024-47554 


DidierLoiseau:
**Describe the bug**
`spring-cloud-openfeign-core` depends on `commons-fileupload:1.5` which in turn depends on  `commons-io:2.11` which suffers from CVE-2024-47554. It would be nice to upgrade it for the next release so that end-users don’t have to force the version of `commons-io`.

Note that Gary D. Gregory indicated in [FILEUPLOAD-357](https://issues.apache.org/jira/browse/FILEUPLOAD-357) that a release of `commons-fileupload:1.6` is already planned with an upgrade `commons-io`, so you may want to just wait for it.

On the other hand, `commons-fileupload` seems to be needed only for `feign-forms-spring`, so maybe both should be made optional? BTW it seems that [`feign-forms-spring` was merged into the main feign project](https://github.com/OpenFeign/feign-form?tab=readme-ov-file#form-encoder), so it will be relocated in its next release (change of `groupId` and alignment on version number).

OlgaMaciaszek:
Hello, @DidierLoiseau, thanks for reporting the issue. https://github.com/OpenFeign/feign-form?tab=readme-ov-file#form-encoder is a separte project, not maintained by our team, and there is no higher version of that project that we could upgrade to. Please create an issue there and link it here so that we can track it and upgrade as soon as it's fixed in the downstream project.

DidierLoiseau:
Thansk @OlgaMaciaszek for your answer. `spring-cloud-openfeign-core` actually [has an exclusion and an explicit dependency on `commons-fileupload`](https://github.com/spring-cloud/spring-cloud-openfeign/blob/ffd0c372e75619ac0c8e9fdbab1d1ae6f0c9ccdc/spring-cloud-openfeign-dependencies/pom.xml#L40-L55), so currently the dependency is maintained here.

I agree that things will be easier when they release a new version with an updated `commons-io`/`commons-fileupload` though, but you will have to deal with the relocation as well then – I just created #1102 for that.

OlgaMaciaszek:
Right @DidierLoiseau; waiting for `commons-fileupload` `1.6` then.

jurgis-sipols:
[...]

Issue #853, opened 2023-03-30
Drop dependency on io.github.openfeign.form:feign-form


vstoyanov:
With the last release 4 years ago and the last accepted pull request about 3 years ago, by all means it is a dead project now. It brings a High Sev CVE to the dependency graph, which is not really nice and certainly not in line with the rest of Spring Framework.

Over the weekend I will do some research on what it could be replaced with and report back.

0. https://central.sonatype.com/artifact/io.github.openfeign.form/feign-form/3.8.0/versions
1. https://github.com/advisories/GHSA-hfrx-6qgj-fp6c

OlgaMaciaszek:
Hello @vstoyanov, thanks for reporting the issue. That dependency is still under official maintenance by the OpenFeign team. A [PR](https://github.com/OpenFeign/feign-form/pull/115) has been submitted to fix the issue. We will monitor what is happening to that PR and try to offer a workaround if necessary.

mgbardakov:
I'm not sure if this fixes the problem. If I'm not mistaken just changing a dependency version isn't enough. https://devhub.checkmarx.com/cve-details/CVE-2023-24998/ A new config FileUploadBase#setFileCountMax should be enabled also.

OlgaMaciaszek:
Hi @mgbardakov, thanks for bringing this up, it's a good point. However, in order to verify this, I have gone through the code in `feign-form` that uses `commons-fileupload` and, actually, a lower-level API is used there only and the `FileUpload` API that is the entry-point for this CVE is not used, so actually, the code would not even be affected by this vulnerability with the `1.4` dep version.

PR #1103, opened 2024-10-11
fix #1102: feign-form-spring relocated under io.github.openfeign

### LINKED ISSUE #1102: feign-form-spring relocated and released as 13.5 under io.github.openfeign
**Describe the bug**
[`feign-formspring`](https://mvnrepository.com/artifact/io.github.openfeign/feign-form-spring) has been relocated under `io.github.openfeign` as part of Feign 13.5, and is now released together with it and [listed in their bom](https://mvnrepository.com/artifact/io.github.openfeign/feign-bom/13.5). `spring-cloud-openfeign` should thus use the new version now that it upgraded to Feign 13.5 in #1099.

DidierLoiseau:
Leaving the exclusion for `commons-io` for now since it still has vulnerabilities in versions < 2.14 (See also #1098), which is still a transitive dependency of [`feign-form-spring:13.5`](https://mvnrepository.com/artifact/io.github.openfeign/feign-form-spring/13.5) (via [`commons-fileupload:1.5`](https://mvnrepository.com/artifact/commons-fileupload/commons-fileupload/1.5)).

### DIFF
--- a/spring-cloud-openfeign-core/pom.xml
+++ b/spring-cloud-openfeign-core/pom.xml
@@ -91,23 +91,15 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>io.github.openfeign.form</groupId>
+			<groupId>io.github.openfeign</groupId>
 			<artifactId>feign-form-spring</artifactId>
 			<exclusions>
-<!--				Vulnerable in 3.8.0-->
+<!--				Still vulnerable in 13.5-->
 				<exclusion>
 					<groupId>commons-io</groupId>
 					<artifactId>commons-io</artifactId>
 				</exclusion>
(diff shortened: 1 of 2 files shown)


DidierLoiseau:
An alternative could be to leave the explicit dependency on `commons-fileupload`, and let dependabot upgrade it when 1.6 gets released. They also have dependabot on OpenFeign so I suppose we can expect the next release to have 1.6 too (if it gets released soon enough).

Related: #853

OlgaMaciaszek (changes requested):
Thanks, @DidierLoiseau. Looks good, but could you submit these changes against the `4.1.x` branch instead of `main`, so that they also get included in the `2023.0.x` release train?

PR #1113, opened 2024-10-19
extract packages from SpringBootApplication or ComponentScan if EnableFeignClients does not set basePackages or basePackageClasses

diguage:


OlgaMaciaszek:
Thanks @diguage. This is not a behaviour we were planning to change. If you'd like to start a discussion on this, including the team and the other users, please create an issue presenting your rationale.

PR #1126, opened 2024-11-20
Dynamically set the url of FeignClient through config

shog808:
Dynamically set the url of FeignClient:

as known if a service have some nodes, we can set the url of the FeignClient to call the certain node。
now i want set the url of FeignClient dynamically in the class FeignClientsRegistrar.java through the config contents of bootstrap.yml.
as:
appoint
then,  call the service message-api will through the url  192.168.2.39:1201

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,14 @@
 				<groupId>io.spring.javaformat</groupId>
 				<artifactId>spring-javaformat-maven-plugin</artifactId>
 			</plugin>
+			<plugin>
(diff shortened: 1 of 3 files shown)


OlgaMaciaszek:
Hello @shog808, thanks for submitting the PR, however, bootstrap is now considered legacy (https://docs.spring.io/spring-cloud-config/reference/client.html#config-first-bootstrap), so we will not be adding in this feature. Thanks you for the work you put into it. In future, feel free to create an issue first in order for the discussion on including proposed feature in the backlog to take place.

Issue #1134, opened 2024-11-29
IllegalArgumentException: Negated params are not supported


ragnor:
After updating from 4.0.x to 4.1.4 I've noticed that my services cannot startup due to:

> Caused by: java.lang.IllegalArgumentException: Negated params are not supported: !sync
>         at org.springframework.cloud.openfeign.support.SpringMvcContract.parseParams(SpringMvcContract.java:372) ~[spring-cloud-openfeign-core-4.1.4.jar:4.1.4]
>         at org.springframework.cloud.openfeign.support.SpringMvcContract.processAnnotationOnMethod(SpringMvcContract.java:249) ~[spring-cloud-openfeign-core-4.1.4.jar:4.1.4]
>         at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:112) ~[feign-core-13.5.jar:na]
>         at org.springframework.cloud.openfeign.support.SpringMvcContract.parseAndValidateMetadata(SpringMvcContract.java:198) ~[spring-cloud-openfeign-core-4.1.4.jar:4.1.4]
>         at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:71) ~[feign-core-13.5.jar:na]
>         at feign.ReflectiveFeign$ParseHandlersByName.apply(ReflectiveFeign.java:140) ~[feign-core-13.5.jar:na]
>         at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:59) ~[feign-core-13.5.jar:na]
>         at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:51) ~[feign-core-13.5.jar:na]
>         at feign.Feign$Builder.target(Feign.java:202) ~[feign-core-13.5.jar:na]
>         at org.springframework.cloud.openfeign.DefaultTargeter.target(DefaultTargeter.java:30) ~[spring-cloud-openfeign-core-4.1.4.jar:4.1.4]
>         at org.springframework.cloud.openfeign.FeignClientFactoryBean.getTarget(FeignClientFactoryBean.java:504) ~[spring-cloud-openfeign-core-4.1.4.jar:4.1.4]
>         at org.springframework.cloud.openfeign.FeignClientFactoryBean.getObject(FeignClientFactoryBean.java:456) ~[spring-cloud-openfeign-core-4.1.4.jar:4.1.4]
[...]

PR #1171, opened 2025-03-09
Add support for negated parameters in SpringMvcContract

kssumin:
This PR adds support for negated parameters in Spring MVC mappings when used with Feign clients. 
It resolves issue #1134 where negated parameters in `@RequestMapping` were not supported, causing `IllegalArgumentException` when shared interfaces were used between server and client.

## Changes:
- Added `allowNegatedParams` configuration option to `FeignClientProperties`
- Modified `SpringMvcContract` to silently skip negated parameters when this feature is enabled
- Added tests for both enabled and disabled scenarios
- Updated documentation to explain the new feature and configuration options

## Use case:
When sharing interfaces between server and client code, negated parameters (e.g. `params = {"!expiration"}`) are useful on the server side to disambiguate mappings. With this change, clients can now use these shared interfaces by setting `spring.cloud.openfeign.client.config.feignName.allowNegatedParams=true`.

The implementation silently ignores negated parameters when building the client request, which aligns with the expected behavior since negated parameters are constraints used only on the server side.

@OlgaMaciaszek

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2023 the original author or authors.
+ * Copyright 2013-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -61,6 +61,11 @@ public class FeignClientProperties {
 	 */
 	private boolean decodeSlash = true;
 
+	/**
+	 * If {@code true}, trailing slashes at the end of request urls will be removed.
+	 */
+	private boolean removeTrailingSlash;
+
 	public boolean isDefaultToProperties() {
 		return defaultToProperties;
 	}
@@ -93,6 +98,14 @@ public class FeignClientProperties {
 		this.decodeSlash = decodeSlash;
 	}
 
+	public boolean isRemoveTrailingSlash() {
+		return removeTrailingSlash;
+	}
+
+	public void setRemoveTrailingSlash(boolean removeTrailingSlash) {
+		this.removeTrailingSlash = removeTrailingSlash;
+	}
+
 	@Override
 	public boolean equals(Object o) {
 		if (this == o) {
@@ -103,12 +116,13 @@ public class FeignClientProperties {
 		}
 		FeignClientProperties that = (FeignClientProperties) o;
 		return defaultToProperties == that.defaultToProperties && Objects.equals(defaultConfig, that.defaultConfig)
-				&& Objects.equals(config, that.config) && Objects.equals(decodeSlash, that.decodeSlash);
+				&& Objects.equals(config, that.config) && Objects.equals(decodeSlash, that.decodeSlash)
+				&& Objects.equals(removeTrailingSlash, that.removeTrailingSlash);
 	}
 
 	@Override
 	public int hashCode() {
-		return Objects.hash(defaultToProperties, defaultConfig, config, decodeSlash);
+		return Objects.hash(defaultToProperties, defaultConfig, config, decodeSlash, removeTrailingSlash);
 	}
 
 	/**
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
@@ -45,6 +45,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  * @author Hyeonmin Park
  * @author Jasbir Singh
  * @author Dominique Villard
+ * @author kssumin
  */
 @ConfigurationProperties("spring.cloud.openfeign.client")
 public class FeignClientProperties {
@@ -65,6 +66,11 @@ public class FeignClientProperties {
 	 * If {@code true}, trailing slashes at the end of request urls will be removed.
 	 */
 	private boolean removeTrailingSlash;
+
+	/**
+	 * If {@code true}, negated parameters (those starting with '!') will be allowed.
+	 */
+	private boolean allowNegatedParams = false;
 
 	public boolean isDefaultToProperties() {
 		return defaultToProperties;
@@ -104,6 +110,14 @@ public class FeignClientProperties {
 
 	public void setRemoveTrailingSlash(boolean removeTrailingSlash) {
 		this.removeTrailingSlash = removeTrailingSlash;
+	}
+
+	public void setAllowNegatedParams(Boolean allowNegatedParams) {
+		this.allowNegatedParams = allowNegatedParams;
+	}
+
+	public boolean getAllowNegatedParams() {
+		return allowNegatedParams;
 	}
 
 	@Override
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
@@ -89,6 +89,7 @@ import static org.springframework.core.annotation.AnnotatedElementUtils.findMerg
  * @author Ram Anaswara
  * @author Sam Kruglov
  * @author Tang Xiong
+ * @author kssumin
  */
 public class SpringMvcContract extends Contract.BaseContract implements ResourceLoaderAware {
 
@@ -117,6 +118,8 @@ public class SpringMvcContract extends Contract.BaseContract implements Resource
 	private final boolean decodeSlash;
 
 	private final boolean removeTrailingSlash;
+
+	private final boolean allowNegatedParams;
 
 	public SpringMvcContract() {
 		this(Collections.emptyList());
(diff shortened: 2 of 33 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
```diff
@@ -172,6 +175,25 @@ public SpringMvcContract(List<AnnotatedParameterProcessor> annotatedParameterPro
...
+
+		annotatedArgumentProcessors = toAnnotatedArgumentProcessorMap(processors);
+		this.conversionService = conversionService;
+		convertingExpanderFactory = new ConvertingExpanderFactory(conversionService);
+		this.decodeSlash = decodeSlash;
+		this.removeTrailingSlash = removeTrailingSlash;
+		this.allowNegatedParams = allowNegatedParams;
 	}
 
 	public SpringMvcContract(List<AnnotatedParameterProcessor> annotatedParameterProcessors,
```
OlgaMaciaszek, line 199: You should set the `allowNegatedParams` field from properties here.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
```diff
@@ -45,6 +45,7 @@
  * @author Hyeonmin Park
  * @author Jasbir Singh
  * @author Dominique Villard
+ * @author kssumin
```
OlgaMaciaszek, line 48: Please use your full name and surname with the author tag.

OlgaMaciaszek, line 48: Please update the date in the license comment to `-2025` in any classes you've modified or added.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
```diff
@@ -106,6 +112,14 @@ public void setRemoveTrailingSlash(boolean removeTrailingSlash) {
 		this.removeTrailingSlash = removeTrailingSlash;
 	}
 
+	public void setAllowNegatedParams(Boolean allowNegatedParams) {
```
OlgaMaciaszek, line 115: let's use `boolean` and not `Boolean` as we do not expect a `null` here.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
```diff
@@ -89,6 +89,7 @@
  * @author Ram Anaswara
  * @author Sam Kruglov
  * @author Tang Xiong
+ * @author kssumin
```
OlgaMaciaszek, line 92: Please use your full name and surname with the author tag.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
```diff
@@ -411,7 +433,11 @@ private void parseParams(MethodMetadata data, Method method, RequestMapping meth
 				data.template().query(resolve(nameValueResolver.getName()), resolve(nameValueResolver.getValue()));
 			}
 			else {
-				throw new IllegalArgumentException("Negated params are not supported: " + param);
+				// Negated parameter case
+				if (!this.allowNegatedParams) {
```
OlgaMaciaszek, line 437: Please remove unnecessary `this` keyword.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientProperties.java
```diff
@@ -66,6 +67,11 @@ public class FeignClientProperties {
 	 */
 	private boolean removeTrailingSlash;
 
+	/**
+	 * If {@code true}, negated parameters (those starting with '!') will be allowed.
```
OlgaMaciaszek, line 71: Please remove "(those starting with '!') " from the comment.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/SpringMvcContractTests.java
```diff
@@ -84,6 +84,7 @@
  * @author Sam Kruglov
  * @author Bhavya Agrawal
  * @author Tang Xiong
+ * @author kssumin
```
OlgaMaciaszek, line 87: Please use your full name and surname with the author tag.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/SpringMvcContractTests.java
```diff
@@ -769,6 +770,28 @@ void shouldSetPageableAsBodyWhenQueryMapParamPresent() {
 		assertThat(data.get(1).bodyIndex()).isEqualTo(0);
 	}
 
+	@Test
+	void testAllowNegatedParams() throws Exception {
```
OlgaMaciaszek, line 774: Please use the same naming convention as other tests in this class, starting with `should...`

OlgaMaciaszek, line 774: Please remove `throws Exception`

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/SpringMvcContractTests.java
```diff
@@ -769,6 +770,28 @@ void shouldSetPageableAsBodyWhenQueryMapParamPresent() {
...
+
+		Method method = TestTemplate_ParseParams.class.getDeclaredMethod("notEqualParams");
+		MethodMetadata data = contract.parseAndValidateMetadata(method.getDeclaringClass(), method);
+
+		assertThat(data.template().url()).isEqualTo("/test");
+		assertThat(data.template().method()).isEqualTo("GET");
+	}
+
+	@Test
+	void testDisallowNegatedParams() throws Exception {
```
OlgaMaciaszek, line 785: Please use the same naming convention as other tests in this class, starting with `should...`,

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/SpringMvcContractTests.java
```diff
@@ -769,6 +770,28 @@ void shouldSetPageableAsBodyWhenQueryMapParamPresent() {
...
+		contract = new SpringMvcContract(Collections.emptyList(), getConversionService(), true, false, false);
```
OlgaMaciaszek, line 786: Please remove `throws Exception`,

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/SpringMvcContractTests.java
```diff
@@ -769,6 +770,28 @@ void shouldSetPageableAsBodyWhenQueryMapParamPresent() {
 		assertThat(data.get(1).bodyIndex()).isEqualTo(0);
 	}
 
+	@Test
```
OlgaMaciaszek, line 773: Please add a test where properties are used to change this. This kind of test could be added, for example, to the `FeignClientFactoryBeanIntegrationTests` class.

OlgaMaciaszek (changes requested):
Hello @kssumin, thanks for submitting the PR. Have added some comments - please address them. Also, please re-submit your PR against `4.1.x` instead of `main`, so that we can include it in earlier supported release trains as well.

OlgaMaciaszek:
Fixes gh-1134.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/support/SpringMvcContractTests.java
kssumin: Regarding your feedback to remove throws Exception from the test methods - I need to handle the exception that comes from the getDeclaredMethod() call:
```java
Method method = TestTemplate_ParseParams.class.getDeclaredMethod("notEqualParams");
```

This method throws NoSuchMethodException which is a checked exception, so I need to either:

1. Declare it specifically with throws NoSuchMethodException
2. Use the broader throws Exception that other tests in this class are using

I noticed other test methods in this class like testProcessAnnotations_ParseParams_SingleParam also use throws Exception, so I wasn't sure if I should keep consistency with existing code or follow a different pattern. What's your preference in this case?"

OlgaMaciaszek: That's ok.

OlgaMaciaszek:
@kssumin the base branch cannot be just changed. The easiest way to correctly change the branch would be to close this PR and create a new one by applying your commits to the `4.1.x` branch.

kssumin:
> @kssumin the base branch cannot be just changed. The easiest way to correctly change the branch would be to close this PR and create a new one by applying your commits to the `4.1.x` branch.

OlgaMaciaszek:
Hello @kssumin, actually I'd really like to ensure that we get this fix into today's `2024.0.1` release, so I'm going to cherry-pick your commits and finish it up quickly. Thank you for working on this issue. I'm going to remove your `@author` tag, since we only accept full name and surname. Please provide your full name and surname and then I will add your `@author` tag to the modified classes.

OlgaMaciaszek:
@kssumin, actually, upon further consideration, since this issue is a bug and not an enhancement, I think we should stop throwing the exception altogether - just add some logging and ignore the parameter on Feign contract processing, so I will add that change now instead. While we'll not be including the proposed changes at this point, seeing your PR has helped in deciding on the fix for the issue, so thank you for your work on this issue.

kssumin:
> Hello @kssumin, actually I'd really like to ensure that we get this fix into today's `2024.0.1` release, so I'm going to cherry-pick your commits and finish it up quickly. Thank you for working on this issue. I'm going to remove your `@author` tag, since we only accept full name and surname. Please provide your full name and surname and then I will add your `@author` tag to the modified classes.

My full name is Kim Su-min. My surname is Kim.

> @kssumin, actually, upon further consideration, since this issue is a bug and not an enhancement, I think we should stop throwing the exception altogether - just add some logging and ignore the parameter on Feign contract processing, so I will add that change now instead. While we'll not be including the proposed changes at this point, seeing your PR has helped in deciding on the fix for the issue, so thank you for your work on this issue.

Thank you for the update. I understand your decision to handle this issue with a logging approach. I'm glad my PR helped in finding the direction for solving the problem. I look forward to actively contributing to the project in the future.

PR #1179, opened 2025-03-31
Replace deprecated URL constructor in getUrl

HyeongDo-Myeong:
The constructor 'URL(String)' has been deprecated since Java 20. Although this does not affect the current environment (Java 17), it will cause issues when upgrading to Java 21 or later. This commit updates the URL validation in FeignClientsRegistrar.getUrl by replacing `new URL(url)` with `new URI(url)`, ensuring forward compatibility. 

Additionally, tests were added to verify both valid and malformed URL scenarios:
- Good URL: "https://good.url"
- Bad URL: "http://bad url"

Verified with Java 17 and the FeignClientsRegistrarTests.


OlgaMaciaszek (changes requested):
Hello @HyeongDo-Myeong, thanks for submitting the PR. Looks good. Please update the date in the license comments on top of all the java classes you've changed to -2025. Also, please add the @author tag with your full name and surname to the javadocs of the classes you've modified.

HyeongDo-Myeong:
@OlgaMaciaszek Thank you for the review. I've updated the license headers to reflect 2025 and added the author tags as requested.

Issue #1191, opened 2025-05-07
Suggestion: Warn or fail when GET method in @FeignClient has unannotated parameters


dkswnkk:
### Summary

This issue is related to the unexpected behavior where a `GET` method in `@FeignClient` silently becomes a `POST` if its parameters are not annotated (e.g., missing `@RequestParam` or `@SpringQueryMap`).

I previously documented the full behavior and reasoning in an issue in the core Feign repository:
https://github.com/OpenFeign/feign/issues/2872

### Clarification

I’ve seen related discussions such as #832, which suggests workarounds like switching the HTTP client to Apache HC5, or adding `@SpringQueryMap`. That may solve specific symptoms, but my suggestion is **preventive** and aimed at improving developer experience.

### Proposal

when a `GET` method contains one or more parameters without annotations like `@RequestParam`, `@PathVariable`, `@SpringQueryMap`, etc., it would be very helpful to:

- Emit a **warning** during startup (default behavior), or
- Optionally **fail fast**, via a configurable flag (e.g., `failOnUnannotatedGetParams = true`)

This could save developers hours of debugging, especially since:
- Feign logs the request as a GET on the client side,
- But the server actually receives a POST request due to how `HttpURLConnection` or other HTTP clients behave with GET + body,
- And the request body is silently empty (`Content-Length: 0`), which causes confusing behavior on the receiving end.

On the sending side, application logs indicate that a GET request was made, but on the receiving side, the request is logged and handled as a POST.
This mismatch makes the issue especially difficult to detect and diagnose.

### Why I'm Opening This Issue Here

Although I originally opened this issue in the core [https://github.com/OpenFeign/feign/issues/2872](https://github.com/OpenFeign/feign/issues/2872), I later looked into where this behavior might originate.

While I’m not 100% certain, it appears that the logic responsible for interpreting annotations like `@RequestParam` and deciding how parameters are bound resides in `SpringMvcContract.java`, which is part of the `spring-cloud-openfeign` project — not the core Feign module.

So I’m opening this issue here as well, in case this is indeed the correct place to address the problem.

OlgaMaciaszek:
Hello @dkswnkk, thanks for reporting the issue. As this project is now in maintenance mode, we only handle dependency upgrades, security issues and bugs. However, feel free to provide a pull request with this change - we'll review it.

dkswnkk:
Hi, I’ve opened a PR for this — appreciate your consideration when convenient.
[...]

PR #1194, opened 2025-05-09
feat: warn when GET method has unannotated parameters

dkswnkk:
## Summary

This PR addresses issue [#1191](https://github.com/spring-cloud/spring-cloud-openfeign/issues/1191) by adding a warning when a `@FeignClient` method is declared as `GET` but has one or more parameters that are **not annotated** with any supported HTTP binding annotations (e.g., `@RequestParam`, `@RequestHeader`, `@PathVariable`, `@SpringQueryMap`, etc.).

## Context

When no annotations are present, Feign may attempt to treat those parameters as a request body. For `GET` methods, this leads to a body-less `POST` due to underlying HTTP client behavior (e.g., `HttpURLConnection`), which causes a discrepancy:

- The client logs show the request as `GET`
- The server receives the request as a `POST`
- No request body is actually sent (`Content-Length: 0`)

This is difficult to debug and confusing in real-world applications.

Related core issue: [OpenFeign/feign#2872](https://github.com/OpenFeign/feign/issues/2872)

## Implementation Details

- Added logic in `SpringMvcContract#parseAndValidateMetadata()` to check for:
  - Method is declared as `GET`
  - Method has parameters
  - None of the parameters have HTTP-related annotations
- If so, a warning is logged via SLF4J:
```java
[OpenFeign Warning] Feign method 'public abstract Response com.example.Client.getSomething(java.lang.String)' is declared as GET with parameters, but none of the parameters are annotated (e.g. @RequestParam, @RequestHeader, @PathVariable, etc). This may result in fallback to POST at runtime. Consider explicitly annotating parameters.
```

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
@@ -68,6 +68,8 @@ import org.springframework.http.InvalidMediaTypeException;
 import org.springframework.http.MediaType;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
@@ -89,6 +91,7 @@ import static org.springframework.core.annotation.AnnotatedElementUtils.findMerg
  * @author Ram Anaswara
  * @author Sam Kruglov
  * @author Tang Xiong
+ * @author Juhyeong An
  */
 public class SpringMvcContract extends Contract.BaseContract implements ResourceLoaderAware {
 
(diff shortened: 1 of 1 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
```diff
@@ -234,7 +237,38 @@ protected void processAnnotationOnClass(MethodMetadata data, Class<?> clz) {
 	@Override
 	public MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method method) {
 		processedMethods.put(Feign.configKey(targetType, method), method);
-		return super.parseAndValidateMetadata(targetType, method);
+		MethodMetadata metadata = super.parseAndValidateMetadata(targetType, method);
+
+		if (isGetMethod(metadata) && method.getParameterCount() > 0 && !hasHttpAnnotations(method)) {
+			LOG.warn(String.format(
```
OlgaMaciaszek, line 243: Please use `if (LOG.isWarnEnabled())[...]`

OlgaMaciaszek (changes requested):
Hello @dkswnkk, thanks for submitting the PR. Have added a comment. Please address it.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
dkswnkk: Thank you for your feedback.
I've updated the code to include the LOG.isWarnEnabled() check as you suggested.
Let me know if you have any further comments!

PR #1207, opened 2025-06-10
Update spring-cloud-commons link

ngocnhan-tran1996:


OlgaMaciaszek (commented):
Hello, @ngocnhan-tran1996, thanks for the PR. Looks good, but could you submit it against `4.2.x` instead of `main`? Like this, it will also be included in earlier release trains.

OlgaMaciaszek (approved):
Thanks @ngocnhan-tran1996. LGTM.

PR #1300, opened 2025-12-09
Fix incorrect HttpHeaders handling in RequestHeaderParameterProcessor

alex-cova:
In Spring 4, HttpHeaders no longer implements the Map interface, which makes the comparison in RequestHeaderParameterProcessor invalid. This PR addresses this issue.


ryanjbaxter:
Please sign your commit(s) so the DCO check passes as well

Issue #1304, opened 2025-12-17
Allow Customizer for `HttpClientConnectionManager` in `HttpClient5FeignConfiguration`


kistlers:
**Is your feature request related to a problem? Please describe.**
The underlying http client (in my case the apache hc5) is already well-customizable via properties and also via supplying my own beans. I'd like a further middle ground by allowing a list of customzer for `HttpClientConnectionManager` in `HttpClient5FeignConfiguration`, as it also exists for the `CloseableHttpClient` in the same file.
In my case, I need to configure the `TlsConfig::handshakeTimeout` as a result of the new handling in 5.5.1 (see https://issues.apache.org/jira/browse/HTTPCLIENT-2386)

**Describe the solution you'd like**
Allow a list of `ObjectProvider<List<HttpClientConnectionManagerBuilderCustomizer>> customizerProvider` to the `hc5ConnectionManager` bean method.

**Describe alternatives you've considered**
Yes, I know I can override the full `HttpClientConnectionManager` as my own bean, but then I loose the rest of the auto-configuration logic for this bean. I can also not just copy the bean metho and add my own logic, as I would also have to copy private methods like `httpsSSLConnectionSocketFactory` and my implementation would, over time, surely diverge for the curated defaults that I would like to keep and profit from.

**Additional context**
I am happy to create a PR, unless there is a good reason why this feature does and should not exist.

PR #1305, opened 2025-12-17
Allow Customizer for HttpClientConnectionManager in HttpClient5FeignConfiguration

kistlers:
The underlying http client (in my case the apache hc5) is already well-customizable via properties and also via supplying my own beans. I'd like a further middle ground by allowing a list of customzer for `HttpClientConnectionManager` in `HttpClient5FeignConfiguration`, as it also exists for the `CloseableHttpClient` in the same file.
In my case, I need to configure the `TlsConfig::handshakeTimeout` as a result of the new handling in 5.5.1 (see https://issues.apache.org/jira/browse/HTTPCLIENT-2386)

Allow a list of `ObjectProvider<List<HttpClientConnectionManagerBuilderCustomizer>> customizerProvider` to the `hc5ConnectionManager` bean method.

Yes, I know I can override the full `HttpClientConnectionManager` as my own bean, but then I loose the rest of the auto-configuration logic for this bean. I can also not just copy the bean metho and add my own logic, as I would also have to copy private methods like `httpsSSLConnectionSocketFactory` and my implementation would, over time, surely diverge for the curated defaults that I would like to keep and profit from.

**Additional context**
I am happy to create a PR, unless there is a good reason why this feature does and should not exist.

This PR implements #1304

Edit: copied over issue description from #1304

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/clientconfig/HttpClient5FeignConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/clientconfig/HttpClient5FeignConfiguration.java
@@ -179,8 +179,8 @@ public class HttpClient5FeignConfiguration {
 	}
 
 	/**
-	 * Callback interface that customize {@link HttpClientBuilder} objects before
-	 * HttpClientConnectionManager created.
+	 * Callback interface that customizes {@link PoolingHttpClientConnectionManagerBuilder}
+	 * objects before HttpClientConnectionManager is created.
 	 *
 	 * @author Severin Kistler
 	 * @since 5.1.0


docs/modules/ROOT/pages/spring-cloud-openfeign.adoc
```diff
@@ -144,7 +144,7 @@ When it comes to the Apache HttpClient 5-backed Feign clients, it's enough to en
 You can customize the HTTP client used by providing a bean of either `org.apache.hc.client5.http.impl.classic.CloseableHttpClient` when using Apache HC5.
 
 You can further customise http clients by setting values in the `spring.cloud.openfeign.httpclient.xxx` properties. The ones prefixed just with `httpclient` will work for all the clients, the ones prefixed with `httpclient.hc5` to Apache HttpClient 5, and the ones prefixed with `httpclient.http2` to Http2Client. You can find a full list of properties you can customise in the appendix.
-If you can not configure Apache HttpClient 5 by using properties, there is an `HttpClient5FeignConfiguration.HttpClientBuilderCustomizer` interface for programmatic configuration.
+If you can not configure Apache HttpClient 5 by using properties, there is an `HttpClient5FeignConfiguration.HttpClientBuilderCustomizer` interface for programmatic configuration. Similarly, to configure the `HttpClientConnectionManager`, you can use `HttpClient5FeignConfiguration.HttpClientConnectionManagerBuilderCustomizer`.
```
ryanjbaxter: Could you add an example to the documentation on how to use the interfaces?

PR #1311, opened 2026-01-15
Fix #1270: Support dash-case for oauth2.clientRegistrationId

### LINKED ISSUE #1270: The configuration value `spring.cloud.openfeign.oauth2.clientRegistrationId` is only accepted with CamelCase and not with dash-case
**Describe the bug**
Using `spring-cloud-openfeign-core-4.3.0.jar`, if the `application.yml` configuration provides 
```
spring:
  security:
    oauth2:
      client:
        registration:
          fancy:
            client-id: jane.appeo
            client-secret: ${env:SUPER_SECRET}
            provider: fancy-authserver
            authorization-grant-type: client_credentials
        provider:
          fancy-authserver:
            token-uri: http://localhost:9999
  cloud:
    openfeign:
      oauth2:
        enabled: true
        clientRegistrationId: fancy
```
When the configuration `spring.cloud.openfeign.oauth2.clientRegistrationId` is set with CamelCase it is matched correctly, but when using `spring.cloud.openfeign.oauth2.client-registration-id` with dash-case the identifier is not matched when creating the `OAuth2AccessTokenInterceptor` at [`FeignAutoConfiguration#defaultOAuth2AccessTokenInterceptor`](https://github.com/spring-cloud/spring-cloud-openfeign/blob/v4.3.0/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java#L388-L392).

**Sample**
This was found debugging some kind of _test_. I will try provide later a full test with the situation

hun-cloud:
## Description

This PR fixes #1270 by adding support for dash-case (kebab-case) property binding for `spring.cloud.openfeign.oauth2.clientRegistrationId`.

Currently, the OAuth2 configuration only accepts camelCase format (`clientRegistrationId`), which is inconsistent with other Spring Boot properties that support relaxed binding (dash-case, snake_case, etc.).

## Changes

- Created `FeignOAuth2Properties` class with `@ConfigurationProperties` annotation
- Replaced `@Value` injection with `FeignOAuth2Properties` in `FeignAutoConfiguration`
- Enabled relaxed binding for OAuth2 configuration properties

### Before
spring:
  cloud:
    openfeign:
      oauth2:
        clientRegistrationId: fancy      # ✅ Works
        client-registration-id: fancy    # ❌ Does not work

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
@@ -81,11 +81,12 @@ import org.springframework.util.ClassUtils;
  * @author Wojciech Mąka
  * @author Dangzhicairang(小水牛)
  * @author changjin wei(魏昌进)
+ * @author jaehun lee
  */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnClass(Feign.class)
 @EnableConfigurationProperties({ FeignClientProperties.class, FeignHttpClientProperties.class,
-		FeignEncoderProperties.class })
+		FeignEncoderProperties.class, FeignOAuth2Properties.class })
 public class FeignAutoConfiguration {
 
 	private static final Log LOG = LogFactory.getLog(FeignAutoConfiguration.class);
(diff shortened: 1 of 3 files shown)


spencergibb:
Duplicates https://github.com/spring-cloud/spring-cloud-openfeign/pull/1312

PR #1333, opened 2026-02-24
Add defensive warning for multiple Feign Client beans to avoid potential resource exhaustion

98001yash:
This change adds a defensive warning in `FeignClientFactoryBean#getTarget()` 
to detect multiple `Client` beans defined within a Feign client context.

Although Feign clients are typically configured with singleton-scoped 
`Client` beans, misconfiguration (such as defining multiple or non-singleton 
`Client` beans) may lead to multiple underlying HTTP connection pools being 
created, potentially resulting in resource exhaustion.

This update does not modify lifecycle behavior or bean creation logic.
It only adds a warning log to improve visibility and prevent accidental misuse.

Additionally, documentation has been updated to clarify that custom 
`Client` beans should be singleton-scoped.


ryanjbaxter:
Now there are a ton of unnecessary formatting changes, checktyle is failing

PR #1366, opened 2026-04-03
Fix issue #1204

### LINKED ISSUE #1204: Response Compression Not Working Due to Missing Accept-Encoding Header
**Describe the bug**
I'm using the compression feature of the spring-cloud-openfeign library. However, I can see that the response compression is not working because the Accept-Encoding header is not being sent on the client side. Additionally, the FeignAcceptGzipEncodingInterceptor is missing from the requestInterceptors field in the SynchronousMethodHandler class.

Interestingly, when I create the following bean:
```java
@Bean
Client defaultClient() {
    return new Client.Default(null, null);
}
```

The Accept-Encoding header starts being sent, and the FeignAcceptGzipEncodingInterceptor appears in the list of requestInterceptors.

**Sample**
To reproduce the bug, I have added a repository: https://github.com/dekholud/spring-cloud-openfeign-issue

ban-xiu:
Close https://github.com/spring-cloud/spring-cloud-openfeign/issues/1204.

Remove @ConditionalOnBean(Client.class) from FeignAcceptGzipEncodingAutoConfiguration FeignAcceptGzipEncodingAutoConfiguration requires a feign.Client bean in the main context to activate. When no explicit Client bean is registered (e.g., no load balancer on classpath), response compression is silently skipped even with compression.response.enabled=true.

FeignAcceptGzipEncodingInterceptor is a RequestInterceptor that only adds an Accept-Encoding header — it has no dependency on Client. Removing this condition makes it consistent with FeignContentGzipEncodingAutoConfiguration, which never had this restriction.

All Feign client implementations (JDK HttpURLConnection, Apache HC5, Java HttpClient) handle gzip decompression transparently, so there is no risk.

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/encoding/FeignAcceptGzipEncodingAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/encoding/FeignAcceptGzipEncodingAutoConfiguration.java
@@ -16,11 +16,9 @@
 
 package org.springframework.cloud.openfeign.encoding;
 
-import feign.Client;
 import feign.Feign;
 
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
The test appears to be failing

Issue #1371, opened 2026-04-30
FeignHttpMessageConverters#initConvertersIfRequired has a concurrency issues


jasonjiang9527:
**Describe the bug**
when first call a feign method ```concurrent``` after application startup ,exception ```Could not write request: no suitable HttpMessageConverter found for request type ``` will be reported .

**Sample**
```java
@FeignClient(name = "spring4-dep", path = "/api", url = "http://localhost:8080")
public interface Api {
    @PostMapping
    String test(@RequestBody Map map);
}


@SpringBootApplication
@EnableFeignClients
public class Spring4DepApplication {
    public static void main(String[] args) {
        ConfigurableApplicationContext run = SpringApplication.run(Spring4DepApplication.class, args);
        Api bean = run.getBean(Api.class);
        List<CompletableFuture<Void>> futureList = new ArrayList<>();
        for (int i = 0; i < 10; i++) {
            futureList.add(CompletableFuture.runAsync(() -> bean.test(Map.of("", ""))));
        }
        CompletableFuture.allOf(futureList.toArray(new CompletableFuture[0])).join();
    }
}
```

Issue #1307, opened 2025-12-20
FeignHttpMessageConverters concurrency issue


alianman:
**Describe the bug**
When mutiple concurrent threads call the `org.springframework.cloud.openfeign.support.FeignHttpMessageConverters#getConverters` method on a non-initialized class, some non-initialized converters could be returned. This leads to similar exceptions when processing requests/responses. After the class is initialized, no exceptions are thrown.

```
feign.codec.EncodeException: Could not write request: no suitable HttpMessageConverter found for request type [com.MyClass]
	at org.springframework.cloud.openfeign.support.SpringEncoder.encodeWithMessageConverter(SpringEncoder.java:168)
	at org.springframework.cloud.openfeign.support.SpringEncoder.encode(SpringEncoder.java:110)
	at org.springframework.cloud.openfeign.support.PageableSpringEncoder.encode(PageableSpringEncoder.java:105)
	at feign.RequestTemplateFactoryResolver$BuildEncodedTemplateFromArgs.resolve(RequestTemplateFactoryResolver.java:274)
	at feign.RequestTemplateFactoryResolver$BuildTemplateByResolvingArgs.create(RequestTemplateFactoryResolver.java:107)
	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:48)
	at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:104)
	at org.springframework.cloud.openfeign.FeignCachingInvocationHandlerFactory$1.proceed(FeignCachingInvocationHandlerFactory.java:66)
	at org.springframework.cache.interceptor.CacheInterceptor.lambda$invoke$0(CacheInterceptor.java:54)
	at org.springframework.cache.interceptor.CacheAspectSupport.invokeOperation(CacheAspectSupport.java:424)
	at org.springframework.cache.interceptor.CacheAspectSupport.execute(CacheAspectSupport.java:410)
	at org.springframework.cache.interceptor.CacheInterceptor.invoke(CacheInterceptor.java:64)
	at org.springframework.cloud.openfeign.FeignCachingInvocationHandlerFactory.lambda$create$0(FeignCachingInvocationHandlerFactory.java:53)
	at jdk.proxy2/jdk.proxy2.$Proxy166.list(Unknown Source)
	... 189 filtered

```

```
java.lang.IllegalArgumentException: 'messageConverters' must not contain null elements
[...]

PR #1372, opened 2026-04-30
Fix issue#1371

jasonjiang9527:


jasonjiang9527:
Fix  #1371

jasonjiang9527 (commented):
reviewed

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/FeignHttpMessageConverters.java
```diff
@@ -51,16 +53,21 @@ public List<HttpMessageConverter<?>> getConverters() {
 
 	private void initConvertersIfRequired() {
 		if (this.converters == null) {
-			this.converters = new ArrayList<>();
-			HttpMessageConverters.ClientBuilder builder = HttpMessageConverters.forClient();
-			// TODO: allow disabling of registerDefaults
-			builder.registerDefaults();
-			// TODO: check if already added? Howto order?
+			synchronized (this.convertersLock) {
```
pdeneve, line 56: Is there a specific reason to sync on another object than `this`?

jasonjiang9527: all the feign call thread will be blocked if another thread synchronize this but not release

VolkerKruse:
Any news? Will this PR be resolved and  merged? The bug causes a lot of problems with people need to manually fix their projects as workaround (see https://github.com/spring-cloud/spring-cloud-openfeign/issues/1307).

EarthCitizen:
This is a major production issue for us.

VolkerKruse:
@jasonjiang9527 
Any hopes that this ever gets merged? The issues mentioned in https://github.com/spring-cloud/spring-cloud-openfeign/issues/1307 are causing big problems. Each service build with a feign client is in risk to cause trouble on production system in case the described manual workaround is not implemented.

PR #1378, opened 2026-05-27
Fix unconditional overwrite of resolved fallbackFactory in FeignClientsRegistrar

### LINKED ISSUE #1367: fallbackFactory property unconditionally overwritten in eagerlyRegisterFeignClientBeanDefinition
**Describe the bug**

In `FeignClientsRegistrar.eagerlyRegisterFeignClientBeanDefinition()` (Spring Cloud OpenFeign 5.0.2-SNAPSHOT / main branch), the `fallbackFactory` property is set twice:

  1. Lines 239-243: Conditionally set with proper `ClassUtils.resolveClassName()` handling
  2. Line 244: Unconditionally overwritten with the raw attribute value

https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java#L239-L244

  ```java
  Object fallbackFactory = attributes.get("fallbackFactory");
  if (fallbackFactory != null) {
      definition.addPropertyValue("fallbackFactory", fallbackFactory instanceof Class ? fallbackFactory
              : ClassUtils.resolveClassName(fallbackFactory.toString(), null));
  }
  definition.addPropertyValue("fallbackFactory", attributes.get("fallbackFactory")); // always overwrites
```

Line 244 unconditionally overwrites the resolved value from lines 239-243, making the ClassUtils.resolveClassName() logic dead code.

This appears to be unintentional because:
  - fallback in the same method (lines 234-238) uses conditional set only — no unconditional overwrite
  - lazilyRegisterFeignClientBeanDefinition() (lines 288-292) handles fallbackFactory with conditional set only                                                                                                                                              

The fix would be removing line 244 to align with both the fallback handling and the lazy registration path.

I'd be happy to submit a PR with the fix and test if you agree this is a bug.

### PR DESCRIPTION
**Pull request by seonwooj0810**
Fixes gh-1367

## Summary

`FeignClientsRegistrar.eagerlyRegisterFeignClientBeanDefinition` resolves the `fallbackFactory` attribute via `ClassUtils.resolveClassName(...)` when it is supplied as a `String`, but the very next line unconditionally re-sets the same property to the raw attribute value, discarding the resolved `Class` and rendering the resolution dead code:

https://github.com/spring-cloud/spring-cloud-openfeign/blob/main/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java#L239-L244

```java
Object fallbackFactory = attributes.get(\"fallbackFactory\");
if (fallbackFactory != null) {
    definition.addPropertyValue(\"fallbackFactory\", fallbackFactory instanceof Class ? fallbackFactory
            : ClassUtils.resolveClassName(fallbackFactory.toString(), null));
}
definition.addPropertyValue(\"fallbackFactory\", attributes.get(\"fallbackFactory\")); // unconditional overwrite
```

The duplicate, unconditional `addPropertyValue` mirrors a pattern that the same method already gets right for `fallback` (lines 234-238 — single conditional set, no overwrite) and that the lazy registration path also gets right (`lazilyRegisterFeignClientBeanDefinition`, lines 288-292). It looks unintentional.

## Change

Remove the duplicate line so the resolved value survives.

## Test plan

- [x] `./mvnw -pl spring-cloud-openfeign-core -am test -Dtest=FeignClientsRegistrarTests` — 15 tests, 0 failures, 0 errors, 1 skipped (existing `testFallback` and `testFallbackFactory` still pass with the `Class<?>` annotation path).

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsRegistrar.java
@@ -241,7 +241,6 @@ class FeignClientsRegistrar implements ImportBeanDefinitionRegistrar, ResourceLo
 			definition.addPropertyValue("fallbackFactory", fallbackFactory instanceof Class ? fallbackFactory
 					: ClassUtils.resolveClassName(fallbackFactory.toString(), null));
 		}
-		definition.addPropertyValue("fallbackFactory", attributes.get("fallbackFactory"));
 		definition.setAutowireMode(AbstractBeanDefinition.AUTOWIRE_BY_TYPE);
 		definition.addPropertyValue("refreshableClient", isClientRefreshEnabled());
 		String[] qualifiers = getQualifiers(attributes);


ryanjbaxter:
Can you please sign your commit so the DCO passes?

Could you submit the PR against the 4.3.x branch?

seonwooj0810:
Thanks for the review, @ryanjbaxter!

I've rebased the commit onto `4.3.x` and added a `Signed-off-by` so DCO should pass now. The PR base has also been switched from `main` to `4.3.x`. Please take another look when you have a moment.

PR #1390, opened 2026-07-01
Fix circuit breaker Feign builder direct usage

### LINKED ISSUE #1210: No CircuitBreakerNameResolver set in circuitBreakerFeignBuilder Bean
We are trying to use Feign with Resilience4j, but without the `@EnableFeignClients` annotation.
When trying to use the `circuitBreakerFeignBuilder` Bean created in https://github.com/spring-cloud/spring-cloud-openfeign/blob/c53d5565cb5d5c32e73a1b07bf835441a4c51132/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java#L232 to create Feign clients (via the `target` method), the following exception is thrown:

```
java.lang.NullPointerException: Cannot invoke "org.springframework.cloud.openfeign.CircuitBreakerNameResolver.resolveCircuitBreakerName(String, feign.Target, java.lang.reflect.Method)" because "this.circuitBreakerNameResolver" is null
	at org.springframework.cloud.openfeign.FeignCircuitBreakerInvocationHandler.invoke(FeignCircuitBreakerInvocationHandler.java:98)
...
```

Unfortunately, the four `FeignCircuitBreaker.Builder` methods `circuitBreakerFactory`, `feignClientName`, `circuitBreakerGroupEnabled` and `circuitBreakerNameResolver` are not public, therefore it is not possible to set the circuitBreakerNameResolver (without reflection).

It would be nice having either a working `circuitBreakerFeignBuilder` bean or to be able to set the properties manually.

hutiefang76:
Fixes gh-1210.

When the circuit breaker `Feign.Builder` bean is used directly, it currently gets created without the `CircuitBreakerFactory`, group setting, or `CircuitBreakerNameResolver` that the normal Feign client targeter path applies later. That makes the direct builder path fail at invocation time instead of behaving like the regular circuit breaker Feign client path.

This change configures the circuit breaker builder bean with the available factory, group flag, and resolver. It also falls back to the Feign target name when the builder is used outside the named Feign client targeter path.

Tests added for:
- direct `Feign.Builder` bean configuration with circuit breaker enabled
- fallback resolver behavior when no resolver bean is available
- using the target name when `feignClientName` is not set

Tested with:

```
./mvnw -pl spring-cloud-openfeign-core -Dtest=FeignAutoConfigurationTests test
```

### DIFF
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
@@ -184,6 +184,11 @@ public class FeignAutoConfiguration {
 			return new AlphanumericCircuitBreakerNameResolver();
 		}
 
+		/**
+		 * @deprecated in favor of
+		 * {@link #circuitBreakerFeignTargeter(CircuitBreakerFactory, FeignCircuitBreakerProperties, CircuitBreakerNameResolver)}.
+		 */
+		@Deprecated
 		@SuppressWarnings("rawtypes")
 		public Targeter circuitBreakerFeignTargeter(CircuitBreakerFactory circuitBreakerFactory,
 				@Value("${spring.cloud.openfeign.circuitbreaker.group.enabled:false}") boolean circuitBreakerGroupEnabled,
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
@@ -37,6 +37,7 @@ import org.springframework.aot.hint.RuntimeHints;
 import org.springframework.aot.hint.RuntimeHintsRegistrar;
 import org.springframework.aot.hint.TypeReference;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -184,14 +185,22 @@ public class FeignAutoConfiguration {
 		}
 
 		@SuppressWarnings("rawtypes")
+		public Targeter circuitBreakerFeignTargeter(CircuitBreakerFactory circuitBreakerFactory,
+				@Value("${spring.cloud.openfeign.circuitbreaker.group.enabled:false}") boolean circuitBreakerGroupEnabled,
+				CircuitBreakerNameResolver circuitBreakerNameResolver) {
+			return new FeignCircuitBreakerTargeter(circuitBreakerFactory, circuitBreakerGroupEnabled,
+					circuitBreakerNameResolver);
+		}
+
+		@SuppressWarnings("rawtypes")
 		@Bean
 		@ConditionalOnMissingBean
 		@ConditionalOnBean(CircuitBreakerFactory.class)
 		public Targeter circuitBreakerFeignTargeter(CircuitBreakerFactory circuitBreakerFactory,
 				FeignCircuitBreakerProperties circuitBreakerProperties,
 				CircuitBreakerNameResolver circuitBreakerNameResolver) {
-			return new FeignCircuitBreakerTargeter(circuitBreakerFactory,
-					circuitBreakerProperties.getGroup().isEnabled(), circuitBreakerNameResolver);
+			return circuitBreakerFeignTargeter(circuitBreakerFactory, circuitBreakerProperties.getGroup().isEnabled(),
+					circuitBreakerNameResolver);
 		}
 
 		static class DefaultCircuitBreakerNameResolver implements CircuitBreakerNameResolver {
--- a/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
+++ b/spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
@@ -37,7 +37,6 @@ import org.springframework.aot.hint.RuntimeHints;
 import org.springframework.aot.hint.RuntimeHintsRegistrar;
 import org.springframework.aot.hint.TypeReference;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -86,7 +85,7 @@ import org.springframework.util.ClassUtils;
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnClass(Feign.class)
 @EnableConfigurationProperties({ FeignClientProperties.class, FeignHttpClientProperties.class,
-		FeignEncoderProperties.class, FeignOAuth2Properties.class })
+		FeignEncoderProperties.class, FeignOAuth2Properties.class, FeignCircuitBreakerProperties.class })
 public class FeignAutoConfiguration {
 
 	private static final Log LOG = LogFactory.getLog(FeignAutoConfiguration.class);
@@ -189,10 +188,10 @@ public class FeignAutoConfiguration {
 		@ConditionalOnMissingBean
 		@ConditionalOnBean(CircuitBreakerFactory.class)
 		public Targeter circuitBreakerFeignTargeter(CircuitBreakerFactory circuitBreakerFactory,
-				@Value("${spring.cloud.openfeign.circuitbreaker.group.enabled:false}") boolean circuitBreakerGroupEnabled,
+				FeignCircuitBreakerProperties circuitBreakerProperties,
 				CircuitBreakerNameResolver circuitBreakerNameResolver) {
-			return new FeignCircuitBreakerTargeter(circuitBreakerFactory, circuitBreakerGroupEnabled,
-					circuitBreakerNameResolver);
+			return new FeignCircuitBreakerTargeter(circuitBreakerFactory,
+					circuitBreakerProperties.getGroup().isEnabled(), circuitBreakerNameResolver);
 		}
 
 		static class DefaultCircuitBreakerNameResolver implements CircuitBreakerNameResolver {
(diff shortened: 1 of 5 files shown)


spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -248,8 +249,14 @@ public Feign.Builder defaultFeignBuilder(Retryer retryer) {
...
-		public Feign.Builder circuitBreakerFeignBuilder() {
-			return FeignCircuitBreaker.builder();
+		public Feign.Builder circuitBreakerFeignBuilder(CircuitBreakerFactory circuitBreakerFactory,
+				@Value("${spring.cloud.openfeign.circuitbreaker.group.enabled:false}") boolean circuitBreakerGroupEnabled,
+				ObjectProvider<CircuitBreakerNameResolver> circuitBreakerNameResolver) {
+			return FeignCircuitBreaker.builder()
+				.circuitBreakerFactory(circuitBreakerFactory)
+				.circuitBreakerGroupEnabled(circuitBreakerGroupEnabled)
+				.circuitBreakerNameResolver(circuitBreakerNameResolver
+					.getIfAvailable(() -> (feignClientName, target, method) -> Feign.configKey(target.type(), method)));
```
ryanjbaxter: Technically I believe the fallback name resolver would be AlphanumericCircuitBreakerNameResolver by default if `alphanumeric-ids.enabled has matchIfMissing = true`.  We should probably check that property and use the correct name resolver

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -251,12 +253,20 @@ public Feign.Builder defaultFeignBuilder(Retryer retryer) {
 		@ConditionalOnBean(CircuitBreakerFactory.class)
 		public Feign.Builder circuitBreakerFeignBuilder(CircuitBreakerFactory circuitBreakerFactory,
 				@Value("${spring.cloud.openfeign.circuitbreaker.group.enabled:false}") boolean circuitBreakerGroupEnabled,
+				@Value("${spring.cloud.openfeign.circuitbreaker.alphanumeric-ids.enabled:true}") boolean alphanumericIdsEnabled,
```
ryanjbaxter: Can you introduce a new public method with the new parameter and delegate to that?  I do not want to change the signature of public methods

ryanjbaxter:
Please sign your commits so the DCO passes

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
hutiefang76: The fallback now follows spring.cloud.openfeign.circuitbreaker.alphanumeric-ids.enabled: it uses AlphanumericCircuitBreakerNameResolver when enabled or omitted, and DefaultCircuitBreakerNameResolver when disabled. Both cases are exercised by the direct-builder tests. Verified with the focused module test: 11 tests, 0 failures/errors.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
hutiefang76: The no-argument public bean method is preserved and delegates to the overload that accepts the configuration values. The test verifies that the public no-argument circuitBreakerFeignBuilder method remains present. Verified with the focused module test: 11 tests, 0 failures/errors.

spring-cloud-openfeign-core/src/test/java/org/springframework/cloud/openfeign/FeignAutoConfigurationTests.java
```diff
@@ -91,6 +99,63 @@ void shouldInstantiateFeignCircuitBreakerTargeterWhenEnabledWithCustomCircuitBre
...
+			.withPropertyValues("spring.cloud.openfeign.circuitbreaker.enabled=true",
+					"spring.cloud.openfeign.httpclient.hc5.enabled=false")
+			.withBean(CircuitBreakerFactory.class, () -> circuitBreakerFactory)
+			.run(ctx -> {
+				Feign.Builder builder = ctx.getBean(Feign.Builder.class);
+
+				assertThat(builder).isInstanceOf(FeignCircuitBreaker.Builder.class)
+					.hasFieldOrPropertyWithValue("circuitBreakerFactory", circuitBreakerFactory)
+					.hasFieldOrProperty("circuitBreakerNameResolver");
+			});
```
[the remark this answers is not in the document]
hutiefang76: The current head verifies the fallback behavior directly. The direct builder test obtains the resolver from the builder, asserts the default AlphanumericCircuitBreakerNameResolver, and invokes it with a real target and method. The disabled-property variant verifies DefaultCircuitBreakerNameResolver and the unmodified config key. Verified with the focused module test: 11 tests, 0 failures/errors.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -249,7 +264,25 @@ public Feign.Builder defaultFeignBuilder(Retryer retryer) {
 		@ConditionalOnMissingBean
 		@ConditionalOnBean(CircuitBreakerFactory.class)
 		public Feign.Builder circuitBreakerFeignBuilder() {
-			return FeignCircuitBreaker.builder();
+			return circuitBreakerFeignBuilder(circuitBreakerFactory, circuitBreakerGroupEnabled, alphanumericIdsEnabled,
+					circuitBreakerNameResolver);
+		}
+
+		public Feign.Builder circuitBreakerFeignBuilder(CircuitBreakerFactory circuitBreakerFactory,
```
ryanjbaxter, line 262: Remove 
```
        @Bean
		@Scope("prototype")
		@ConditionalOnMissingBean
		@ConditionalOnBean(CircuitBreakerFactory.class)
``` 
From the old method, add them here and then inject the beans you need instead of autowiring them

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
```diff
@@ -237,6 +240,18 @@ public Feign.Builder feignBuilder(Retryer retryer) {
 	@ConditionalOnProperty("spring.cloud.openfeign.circuitbreaker.enabled")
 	protected static class CircuitBreakerPresentFeignBuilderConfiguration {
 
+		@Autowired
+		private CircuitBreakerFactory circuitBreakerFactory;
+
+		@Value("${spring.cloud.openfeign.circuitbreaker.group.enabled:false}")
```
ryanjbaxter: I think we should probably create a Properties class for these, the group property is already used in `FeignAutoConfiguration`

ryanjbaxter (changes requested):

ryanjbaxter:
Can you rebase the PR?

hutiefang76:
@ryanjbaxter Rebased onto the current `main` and addressed the requested configuration cleanup.

- Added `FeignCircuitBreakerProperties` for the existing `group.enabled` and `alphanumeric-ids.enabled` settings, shared by the targeter and builder.
- Moved the prototype builder bean annotations to the injected overload; the existing public no-arg method remains available.
- Removed field injection from the circuit-breaker builder configuration.
- Extended the tests to assert the actual injected bean method and both resolver behaviors.

Validation: `./mvnw -pl spring-cloud-openfeign-core -Dtest=FeignAutoConfigurationTests test` passes (11 tests). I also ran the full core module; its two remaining failures are in `FeignClientCacheTests`, which expects `UnknownHostException` for the external fake host `foo`, while this local network resolves it and returns `NoHttpResponseException`.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
```diff
@@ -189,10 +188,10 @@ public CircuitBreakerNameResolver alphanumericCircuitBreakerNameResolver() {
 		@ConditionalOnMissingBean
 		@ConditionalOnBean(CircuitBreakerFactory.class)
 		public Targeter circuitBreakerFeignTargeter(CircuitBreakerFactory circuitBreakerFactory,
```
ryanjbaxter, line 204: Same overload pattern needs to be applied here

ryanjbaxter (changes requested):
PR build passes, one small change

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignAutoConfiguration.java
hutiefang76: Applied the same overload pattern here: restored the public circuitBreakerFeignTargeter(CircuitBreakerFactory, boolean, CircuitBreakerNameResolver) API and added the property-injected @Bean overload that delegates to it. Added a reflection-based compatibility test. FeignAutoConfigurationTests passes (12 tests).

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
hutiefang76: Added FeignCircuitBreakerProperties for group.enabled and alphanumeric-ids.enabled, enabled it in FeignAutoConfiguration, and inject it into the managed builder and targeter overloads. The legacy public targeter overload retains its boolean parameter for API compatibility; the managed overload delegates with the bound property value.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientsConfiguration.java
hutiefang76: Moved the bean and conditional annotations to the injected overload. The original no-argument builder method remains a public, unannotated API, while the managed overload receives the dependencies. FeignAutoConfigurationTests passes (12 tests).

ryanjbaxter (changes requested):
One small thing and I think this is good to go.

We should add an `@Deprecated` annotation on the old auto configuration methods pointing at the new ones.

hutiefang76:
Addressed the compatibility follow-up in 8876ec32:\n\n- marked both legacy public auto-configuration methods as @Deprecated, with Javadoc links to the property-based overloads;\n- retained the old signatures for binary compatibility;\n- added reflection assertions for the deprecation annotations.\n\nVerified with ./mvnw -pl spring-cloud-openfeign-core -Dtest=FeignAutoConfigurationTests test (12 tests passing).

PR #1405, opened 2026-08-21
Resolve placeholders before parsing request headers

yashdotdev13:
## Description

Resolves property placeholders in the complete `RequestMapping.headers` expression before parsing the header name and value.

Previously, placeholders worked when used as part of a header value:

```java
headers = "myHeader=${my.header.value}"


ryanjbaxter:
Was there a change that broke this functionality?

yashdotdev13:
I looked through the history, and I couldn't identify a specific change that broke this functionality.

It appears that placeholder resolution for header names and values has been supported, but resolving a placeholder for the entire RequestMapping.headers expression was not handled. The existing parsing logic looks for = before resolving the expression, so a value such as ${feignClient.placeholderHeader} is treated as a complete header expression without first being resolved.

The existing getHelloHeadersPlaceholders() test has been present since the initial commit, and the history suggests this particular case was not previously supported rather than being broken by a recent change.

My change resolves the header expression before parsing it, allowing placeholders that resolve to a complete name=value header expression while preserving the existing behavior for placeholders used within the header name or value.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/support/SpringMvcContract.java
ryanjbaxter: Are the `resolve` calls here necessary with the new `resolve` call above?

yashdotdev13: Good point, no, those calls are redundant after resolving the complete header expression first. I'll remove them.

yashdotdev13: Removed the redundant resolve calls. The complete header expression is now resolved before it is parsed.

ryanjbaxter:
Could you merge in the lates changes from main into your branch please before I merge this?  Thanks!

yashdotdev13:
Done, thanks! I've merged the latest changes from main into my branch and verified that the tests pass.

PR #1410, opened 2026-09-03
Do not warn about GET parameters bound by a parameter processor

HDPark95:
`hasHttpAnnotations` tested a hardcoded list of five annotations before logging the "declared as GET with parameters, but none of the parameters are annotated" warning.

The contract binds more than those. `@MatrixVariable`, `@RequestPart` and `@CookieValue` have default processors, users can register their own, and an unannotated `Pageable` becomes the query map. None can fall back to POST, yet all warn today, including the docs' `StoreClient` example and `TestTemplate_MatrixVariable` here.

The check now looks the annotation up in `annotatedArgumentProcessors`, keeps feign's `@QueryMap` explicit as it has no processor, and adds `Pageable`, whose `NoClassDefFoundError` guard moves into `isPageable`.

`./mvnw -pl spring-cloud-openfeign-core test` passes on JDK 21; the four new tests fail without the change.


ryanjbaxter:
Cherry picked into 5.0.x as well

PR #1417, opened 2026-09-11
Scope repeated Client resolution warning to the application context

kdelay:
`FeignClientFactoryBean` tracks resolved context ids in a `private static final Set`, so the "is being initialized more than once. Ensure the Client bean is Singleton scoped" warning is JVM-wide and never cleared.

A second `ApplicationContext` in the same JVM, which is normal in a Spring Boot test suite or after `@DirtiesContext`, resolves the `Client` for a context id exactly once and still warns. The advice is wrong there: the bean is singleton scoped.

The detection is unchanged. The set is now per application context, keyed weakly on the owning `BeanFactory` so closed contexts are not retained.

`FeignClientFactoryBeanTests` covers both sides: a second resolution inside one context still warns, two separate contexts do not. The second test fails on `main` and passes here.


ryanjbaxter:
Can you submit this agains the 5.0.x branch?

kdelay:
Retargeted to 5.0.x in fddd1ee. The commit cherry-picked cleanly. On 5.0.x the new test fails without the fix (FeignClientFactoryBeanTests.java:61, the captured log still contains the repeated-initialization warning) and passes with it; ./mvnw -pl spring-cloud-openfeign-core verify gives 443 tests, 0 failures, 0 checkstyle violations.

spring-cloud-openfeign-core/src/main/java/org/springframework/cloud/openfeign/FeignClientFactoryBean.java
```diff
@@ -88,7 +90,13 @@ public class FeignClientFactoryBean
 
 	private static final Log LOG = LogFactory.getLog(FeignClientFactoryBean.class);
 
-	private static final Set<String> resolvedContextIds = Collections.newSetFromMap(new ConcurrentHashMap<>());
+	/**
+	 * Context ids whose {@link Client} bean has already been resolved, tracked per
+	 * application context so that a second application context in the same JVM does not
+	 * look like repeated initialisation.
+	 */
+	private static final Map<Object, Set<String>> resolvedContextIds = new ConcurrentReferenceHashMap<>(2,
```
ryanjbaxter: I think we should do this instead
```
private static final Map<Object, Set<String>> resolvedContextIds = Collections
    .synchronizedMap(new WeakHashMap<>());
```

The reason being I am concerned about garbage collection removing the context ids, see the note in ConcurrentReferenceHashMaps' javadocs
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/ConcurrentReferenceHashMap.html

ryanjbaxter (changes requested):

kdelay:
Switched to Collections.synchronizedMap(new WeakHashMap<>()) in d3ebc2b. Verified on 5.0.x: the new test still fails without the fix (AssertionError at FeignClientFactoryBeanTests.java:61) and passes with it; ./mvnw -pl spring-cloud-openfeign-core verify gives 443 tests, 0 failures, 0 checkstyle violations.

